Issue #1, opened 2015-06-17
Review imported messages from ROS 1


wjwwood:
In 5240924ea4896f2614416bc34ff9ddc0439f1df3 I just copied them over 1 for 1, only removing trailing whitespace and ensuring a newline at the end of the files. You mentioned that there would be some changes that need to be made to the fields? What additional things should I check? Some things I plan to do currently:

- Replace instances of `time` and `duration` with `builtin_interfaces/Time` and `builtin_interfaces/Duration`.
- Clean up comments where appropriate, e.g. Header still refers to conventions about numeric frame_id's.

I'll also propose pr's for some more substantial changes which I think we'll have to discuss before committing.

Is it required to always use `<package_name>/<msg_name>` now? Or is just the `<msg_name>` allowed within the same package? I think it's the former, and if so, I think we'll need to fix that in a few places as well.

@ros2/owners

dirk-thomas:
* The current parser uses relaxed rules in order to handle ROS 1 messages (https://github.com/ros2/rosidl/blob/7fbc60bc0c041796a603a4e301108d0e5887e75e/rosidl_parser/rosidl_parser/__init__.py#L51-L56). When you switch the regular expressions to the strict versions they should still pass.

* Update the rmw implementation to handle the new package name of builtin messages.

* Relative message references are fine, see https://github.com/ros2/design/blob/gh-pages/articles/110_interface_definition.md:

  ```
  The *base type* can be one of the following:
  * a primitive type from the list above: e.g. `int32`
  * a string with an upper boundary: `string<=N` to limit the length of the string to `N` characters
  * a complex type referencing another message:
    * an *absolute* reference of a message: e.g. `some_package/SomeMessage`
    * a *relative* reference of a message within the same package: e.g. `OtherMessage`
  ```

dirk-thomas:
Please make sure to enable the linters for the message packages.

wjwwood:
> The current parser uses relaxed rules in order to handle ROS 1 messages ...

Ok, I'll make a PR with the fixes that those stricter rules require, so everyone can see it.

> Update the rmw implementation to handle the new package name of builtin messages.

I'll do that when I propose a pr to remove `builtin_msgs` from `examples`. Just trying to get these setup before making that change.

> Please make sure to enable the linters for the message packages.

Will do. Is it just the `ament_lint_auto` and `ament_lint_common` dependencies that I need?

wjwwood:
[...]

PR #2, opened 2015-06-17
remove seq number from Header.msg

wjwwood:
This has been proposed as far back as six years ago, but was avoided in ROS 1 to prevent disruption to existing code. The idea is that having a sequence number that is inconsistently set by the middleware, and often set by users, only to have the middleware override it, doesn't really make sense. Rather the middleware should give the sequence number to the subscriber through a separate channel (not as part of the message struct), and prevent the publishing user from setting it manually (it doesn't make sense for them to do so). Therefore it doesn't need to be a field in the `.msg`.

Also, unlike `stamp` and `frame_id`, `seq` is not needed to relate one piece of data to another. Both `stamp` and `frame_id` are values with global significance and are values that the middleware cannot realistically set automatically. Whereas `seq` is specific to a single publishing source, or topic depending on the middleware's behavior, and doesn't help coordinate the message instance with message instances from other sources or topics. It might be valid for messages to have sequence numbers if the source of the sequence numbers is generated by the user rather than the middleware, but is not needed in `Header.msg` interface. Another reason for keeping `stamp` in the header while `seq` should be removed, is that the middleware might provide the subscriber with a message received timestamp, but that is typically distinctly different from the creation of the data timestamp in the `Header.msg`.

Links to previous discussions about this issue are in this comment:

https://github.com/ros2/common_interfaces/issues/1#issuecomment-112621348

Request for comment.

### DIFF
--- a/std_interfaces/msg/Header.msg
+++ b/std_interfaces/msg/Header.msg
@@ -2,9 +2,6 @@
 # This is generally used to communicate timestamped data
 # in a particular coordinate frame.
 
-# Sequence ID: consecutively increasing ID
-uint32 seq
-
 # Two-integer timestamp that is expressed as seconds and nanoseconds.
 builtin_interfaces/Time stamp
 


dirk-thomas:
LGTM

esteve:
+1

tfoote:
+1

tfoote:
To send email to community for comment.

rethink-imcmahon:
+1 I think the rational is solid. Elimination of the inconsistent seq usage is definitely welcome. It was always a guessing game who (pub/sub) was supposed to set it, depending on the package.

jack-oquin:
+1

jacquelinekay:
+1

baalexander:
+1

gerkey:
Lots of +1.  Please merge.

wjwwood:
Ok, thanks for everyone's feedback. If you have positive comments or constructive criticism please feel free to continue the discussion here.

adolfo-rt:
+1 Late, but for the record.

ros-discourse:

https://discourse.ros.org/t/foxy-message-api-review/13105/8

PR #8, opened 2016-02-03
add AMENT_IGNORE files into currently unused msg packages

dirk-thomas:


jacquelinekay:
I'm not sure if we should always ignore these packages. I would instead prefer to script Jenkins to add the file to packages we want to ignore.

dirk-thomas:
We already do exactly the same for:
* https://github.com/ros2/common_interfaces/blob/master/actionlib_msgs/AMENT_IGNORE
* https://github.com/ros2/common_interfaces/blob/master/diagnostics_msgs/AMENT_IGNORE
* https://github.com/ros2/common_interfaces/blob/master/shape_msgs/AMENT_IGNORE
* https://github.com/ros2/common_interfaces/blob/master/trajectory_msgs/AMENT_IGNORE
* https://github.com/ros2/common_interfaces/blob/master/visualization_msgs/AMENT_IGNORE

So I don't see why the new three packages should be any different. And at the time we chose this approach (over blacklisting on the build farm) since also every local build benefits from it.

wjwwood:
Based on the lunch discussion we'll use `AMENT_IGNORE` for permanent ignores like this one and also merge the `--skip-packages` pull request but only to be used to make the CI jobs more flexible.

PR #9, opened 2016-02-03
disable many uncommonly used messages temporarly to save build time

wjwwood:
As we discussed on the mailing list.

The only thing I could think we might do differently is that we could add an option to opt-in or opt-out of building these messages using an environment variable. On the one hand that's a strange (surprising) side effect, but on the other we could re-enable these messages for the nightly/packaging jobs more easily.

### DIFF
--- a/geometry_msgs/CMakeLists.txt
+++ b/geometry_msgs/CMakeLists.txt
@@ -11,29 +11,33 @@ find_package(rosidl_default_generators REQUIRED)
 find_package(std_msgs REQUIRED)
 
 set(msg_files
+# TODO(wjwwood): re-enable all messages when build time is not so important.
+#                This will likely be the case when we have a proper build farm
+#                and/or we reduce the number of middlewares or languages we
+#                build for by default.
   "msg/Point.msg"
   "msg/Point32.msg"
-  "msg/PointStamped.msg"
-  "msg/Polygon.msg"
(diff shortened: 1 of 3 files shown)


wjwwood:
Also, I left a few additional messages in (e.g. LaserScan and PointCloud* messages since I thought we might use them in the near future) and disabled some more in `std_msgs`, since it also has a lot of uncommonly used messages.

dirk-thomas:
Can you please trigger a CI job for this to ensure that it works.

wjwwood:
Yeah I will after I finish the local build.

wjwwood:
Ok, it builds locally for me now, here is CI:

- http://ci.ros2.org/job/ci_linux/889/
- http://ci.ros2.org/job/ci_osx/742/
- http://ci.ros2.org/job/ci_windows/972/

wjwwood:
I had tf2 disabled locally, trying again. I'll post CI links when they are actually passing.

wjwwood:
Passing CI:

- http://ci.ros2.org/job/ci_linux/890/
- http://ci.ros2.org/job/ci_osx/743/
 - It has failing tests but they seem unrelated
- http://ci.ros2.org/job/ci_windows/973/
 - Also has unrelated test failures

Also I squashed the commits.

dirk-thomas:
lgtm

PR #10, opened 2016-02-16
Stop ignoring std_srvs

jacquelinekay:
Reenabling this package because it contains only services, and we recently had a regression for packages of this type. Open to suggestions for other packages to enable for this use cas.e


dirk-thomas:
I think we should keep this package disabled for now. Simply because we disabled it to reduce build time in the first place.

jacquelinekay:
Alright, fine with me.

PR #12, opened 2016-03-29
Add introspection_msgs

vmayoral:
These messages are used in rosnode list functionality.

### DIFF
--- a/introspection_msgs/msg/ROSMeta.msg
+++ b/introspection_msgs/msg/ROSMeta.msg
@@ -1,0 +1,2 @@
+string node_name 	# ROS node name
+uint8[] id 			# DDS ID
--- a/introspection_msgs/CMakeLists.txt
+++ b/introspection_msgs/CMakeLists.txt
@@ -1,0 +1,22 @@
+cmake_minimum_required(VERSION 2.8.3)
+
+project(introspection_msgs)
+
+if(NOT WIN32)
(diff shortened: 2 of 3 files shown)


introspection_msgs/msg/ROSMeta.msg
```diff
@@ -0,0 +1,2 @@
+string node_name 	# ROS node name
+uint8[] id 			# DDS ID
```
tfoote: Can you add some documentation to this message? A high level description would be helpful. And in particular why is the `id` field a vector? The short comment suggests that it's one ID.

And the name is not descriptive. Such that you only get a sense of the fact that it has metadata, but it clearly does not have all the metadata. It looks right now to be a name to id mapping.

wjwwood:
One general comment is that it might make sense for these message to live in the `rcl_interfaces` package.

vmayoral:
Thanks @tfoote and @wjwwood for the comments. 

Representing the `id` as a `string` instead of a `uint8[]` makes sense. With regard the naming and locations, I don't have a strong opinion about it.

I've made an attempt to document a bit more the message at https://github.com/erlerobot/common_interfaces/commit/dcd9fa517c9e008e87943eebff84f6c8be34d728. Feel free to cherry-pick the changes if you find it useful.

mikaelarguedas:
closing this in the favor of the command line tools implemented in the [ros2cli](https://github.com/ros2/ros2cli/) repository. thanks @vmayoral for all your hard work!

PR #15, opened 2016-04-19
Enable some messages

codebot:
These messages are needed for geometry stuff (e.g., eigen_conversions) to compile in the ros12 shim


std_msgs/msg/Empty.msg
```diff
@@ -0,0 +1,3 @@
+# Temporary dummy field to get around the fact that some IDL parsers (e.g.,
+# OpenSplice) don't like having empty messages
+uint8 dummy
```
dirk-thomas: I don't know what you did when you ran into this but you don't have to add any fields to an .msg file to make it work with OpenSplice. This happens internally and is transparent to the user. Other ROS 2 packages do contain empty message definition files without a problem.

gerkey: Indeed, it's not needed! I had a compile error for Empty.msg that I attributed to its emptiness, but I must have been had some other problem at the time. I just tried with it being empty and that worked fine.

PR #18, opened 2016-06-22
fix export of rosidl_default_runtime

dirk-thomas:
Connect to ros2/rosidl#158 and requires that `rosidl_default_runtime` doesn't export all the typesupport packages anymore.


wjwwood:
lgtm

PR #19, opened 2016-06-29
enable some messages

gerkey:
These messages are needed for controlling the TurtleBot.


tfoote:
+1

wjwwood:
+1 (but out of curiosity why do you need nav_msgs?)

dirk-thomas:
Especially all messages and service of `nav_msgs`. Those should be restricted to what is actually needed.

gerkey:
TurtleBot needs nav_msgs/Odometry. I commented out the other ones.

wjwwood:
Thanks @gerkey, I believe it saves us a lot of build time by keeping the messages we aren't using commented out.

mikaelarguedas:
I noticed that we never build/test the turtlebot demos on CI. Should we revert this PR to avoid generating/testing all these "unused" messages?

dirk-thomas:
The question is do we expect anyone to build the turtlebot demo? If no, we can safely remove the message again. Otherwise I think it is more convenient to leave them enabled. We could still consider to not test them anymore?

mikaelarguedas:
That is a good point, on the same idea the people who ported [turtlesim to ros2](https://github.com/sukha-cn/turtlesim-ros2) are using some messages that would end up not being generated anymore if we merge these. Closing them then. I have other ideas on how to reduce build/testing time and will open tickets accordingly

PR #20, opened 2016-11-04
added AMENT_IGNORE

mikaelarguedas:
see https://github.com/ros2/common_interfaces/pull/19#issuecomment-258320001


mikaelarguedas:
see https://github.com/ros2/common_interfaces/pull/19#issuecomment-258578156

PR #24, opened 2016-11-29
add lifecycle_msgs

Karsten1987:


lifecycle_msgs/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+
+find_package(ament_cmake REQUIRED)
+find_package(rosidl_default_generators REQUIRED)
+find_package(std_msgs REQUIRED)
+
+set(msg_files
+  "msg/Transition.msg"
+)
+set(srv_files
+  "srv/GetState.srv"
```
mikaelarguedas: nicktip: I think we aim for alphabetical order

lifecycle_msgs/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+  "srv/ChangeState.srv"
+)
+
+rosidl_generate_interfaces(${PROJECT_NAME}
+  ${msg_files}
+  ${srv_files}
+  DEPENDENCIES std_msgs
```
mikaelarguedas: I don't think that you need to depend on `std_msgs` if you are using only primitive types (dpeendency can be removed from `package.xml` too)

tfoote:
We commonly put the state enumerations into the msg to make sure that they're available cross platform. I know that this is a little bit lower level and the goal is for all of this to be in rcl only but even if we choose not to enumerate here, it would be great to add a comment where to look up the valid states. It might make sense to simply have a State message with the integer and the enums. It's extra overhead but improves insepction. And can be reused for publishing the state and rendering the state in the future since the data will be semantically typed.

dirk-thomas:
Should these messages maybe be closer to the other messages using in rcl (from `rcl_interfaces`)? I would prefer `rcl` to not depend on the `common_interfaces` repo.

Karsten1987:
I'd left out the states names on purpose for now or two reasons:
- optimization. This also goes along with the fact that in-code I can compare my state to the actual enum value. Therefore I can still avoid using the actual state number.
- customization: Might be a minor point here, but in case users wanted to create their own states, they don't have to modify the lifecycle_msgs

Karsten1987:
@dirk-thomas What exactly is the difference between `rcl_interfaces` and `common_interfaces`?  This package contains only `lifecycle_msgs`, that'd be the only dependency.

dirk-thomas:
Repositories will need to be released in a specific order. If rcl(*) uses this new package the whole repo needs to be released before. That is the reason why `rcl_interfaces` is in a separate repo.

lifecycle_msgs/srv/GetState.srv
```diff
@@ -0,0 +1,3 @@
+string node_name
+---
+uint8 current_state
```
gbiggs: What about putting the constants for the default life cycle state in a separate file that gets included by both message definitions?

Karsten1987:
closed in favor of https://github.com/ros2/rcl_interfaces/pull/11

Issue #27, opened 2017-01-19
Could not find the sensor_msgs/BatteryState?


sidd1:
I was not able to locate the sensor_msgs/BatteryState msg anywhere in the ros2 repo. However I did notice that all the other messages from sensor_msgs have been ported. Am I missing something?

mikaelarguedas:
I think the reason is that `BatteryState` has been added to `sensor_msgs` after the initial port of this package to ROS2. I don't see a reason not to port it too. Thanks for pointing it out

sidd1:
Following up, the `Header` message in `std_msgs` does not have the `seq` field. Could be the same reason that it was updated after the port of this package.

codebot:
The DDS/RTPS protocol (and many/most middlewares in general) includes a sequence number for each message for internal flow control, so the idea was the `seq` field in the ROS message header would be redundant. In the RTPS case, there is some text describing the sequence number field on page 51 of the [official spec PDF](http://www.omg.org/spec/DDSI-RTPS/2.2/PDF/), and more details of this field throughout the document. Cheers

mikaelarguedas:
Most of the high level changes between ROS1 and ROS2 are documented [here](http://design.ros2.org/articles/changes.html) if you want to have an overview without diving into the details of the spec

sidd1:
Ah, I missed that. Thanks!

tfoote:
Also note that the seq element has been deprecated since noteably before 2013. It was left for backwards compatibility. The oldest link I can find is here: http://answers.ros.org/question/55126/why-does-ros-overwrite-my-sequence-number/?answer=55132#post-id-55132

wjwwood:
See this discussion: https://groups.google.com/forum/#!topic/ros-sig-ng-ros/_UFMxG5TG9s

PR #28, opened 2017-01-31
port missing interfaces from ROS1 common_msgs

mikaelarguedas:
Connects to https://github.com/ros2/common_interfaces/issues/27
Closes https://github.com/ros2/common_interfaces/issues/27

This PR doesn't include `SetBool.srv` from `std_srvs` to follow the "services names should carry semantic meaning" guideline

### DIFF
--- a/diagnostic_msgs/CMakeLists.txt
+++ b/diagnostic_msgs/CMakeLists.txt
@@ -17,6 +17,7 @@ set(msg_files
 )
 set(srv_files
   "srv/SelfTest.srv"
+  "srv/AddDiagnostics.srv"
 )
 rosidl_generate_interfaces(${PROJECT_NAME}
   ${msg_files}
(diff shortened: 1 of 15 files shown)


dirk-thomas:
I would include `std_srvs/SetBool.srv`. While we don't recommend users to rely on it in their API (the same for a lot of other messages in `std_msgs`) this interface is convenient for prototyping and testing.

mikaelarguedas:
Fair enough https://github.com/ros2/common_interfaces/pull/28/commits/fbad6e210736c220fd7d6650eae0472a50e0176e

I'm keeping this PR in backlog because I'm not sure we want to increase build time for messages we dont use yet

diagnostic_msgs/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ set(msg_files
 )
 set(srv_files
   "srv/SelfTest.srv"
+  "srv/AddDiagnostics.srv"
```
dirk-thomas: Order.

mikaelarguedas:
Good catch, thanks

mikaelarguedas:
@ros2/dev planning on merging this given that I expect `nav_msgs` to be reenabled soon and that the additional build time will not be  significant

Karsten1987:
+1 for enabling nav_msgs

PR #30, opened 2017-03-29
add test_msgs

Karsten1987:
adding a test_msgs package which serves for various testing purposes.

It will replace the mock_msgs inside rclcpp https://github.com/ros2/rclcpp/blob/master/rclcpp/CMakeLists.txt#L149

It will serve as a testing dependency for rmw unit testing as well as cli_tools.

This PR is motivated by https://github.com/ros2/system_tests/issues/191


Karsten1987:
It currently fails on tests:

```
27: /Users/karsten/workspace/osrf/ros2_ws/build_isolated/test_msgs/rosidl_generator_py/test_msgs/msg/_bounded_array_nested.py:54:9: F401 'test_msgs.msg.Primitives' imported but unused
27:         from test_msgs.msg import Primitives
27:         ^
27:
27: 1     F401 'test_msgs.msg.Primitives' imported but unused
27:
27: 21 files checked
27: 1 errors
27:
27: 'F'-type errors: 1

[...]

The following tests FAILED:
	 27 - flake8_rosidl_generated_py (Failed)
```

mikaelarguedas:
Looks good to me so far. Need to looks at the parser and python message generation to see if we can remove unwanted imports

mikaelarguedas:
Looks like the parser doesn't provide a way to dissociate easily if an array is bounded `[<=N]` or static `[N]` (https://github.com/ros2/rosidl/blob/master/rosidl_parser/rosidl_parser/__init__.py#L200) both are `is_array` and have an `upper_bound` so right know we'll have to stick with the linter complain.

dirk-thomas:
> Looks like the parser doesn't provide a way to dissociate easily if an array is bounded [<=N] or static [N] (https://github.com/ros2/rosidl/blob/master/rosidl_parser/rosidl_parser/__init__.py#L200) both are is_array and have an upper_bound so right know we'll have to stick with the linter complain.

For arrays of static size `is_upper_bound` is `False`. For arrays with an upper bound `is_upper_bound` is `True`. See the convenient methods on the class: https://github.com/ros2/rosidl/blob/267d480b078e34d034020de2313a01c47856644b/rosidl_parser/rosidl_parser/__init__.py#L246-L250

mikaelarguedas:
oh you're right! I missed that for some reason, sorry for the noise. Proper fix in python generator: https://github.com/ros2/rosidl/issues/210

mikaelarguedas:
@ros2/team what would be the best way for packages depending on this to get the list of all the interfaces of this package in CMake ? I know we have `ament_index` for the cpp/python use case but do we have any way to do that easily in CMake?
I'm thinking about this because `test_communication` will need something like this to be able to move all its messages in this new package

wjwwood:
For CMake packages, you could just put them in a CMake variable that gets populated in a CMake extras file for this package. Basically that would make the CMake variable available after anyone does `find_package(test_msgs ...)`.

For Python support (`setup.py` only), I suppose you could put them in a `ament_index` file. You can access the `ament_index` with CMake, in addition to C++ and Python, see:

https://github.com/ament/ament_cmake/blob/master/ament_cmake_core/cmake/index/ament_index_get_resource.cmake

mikaelarguedas:
oh cool I didn't know about that. Thanks @wjwwood for the pointer!

mikaelarguedas:
@Karsten1987 while being ready for merging I'd rather not merge this before we have the corresponding PR on test_communication to leverage this (otherwise we will generate these messages twice without taking advantage of this new package)

mikaelarguedas:
Just occured to me that maybe this should go in the `rcl_interfaces` repo so that client libraries can depend on it for testing without depending on the `common_interfaces` repo. @ros2/team any thoughts about this ?

wjwwood:
> Just occured to me that maybe this should go in the rcl_interfaces repo so that client libraries can depend on it for testing without depending on the common_interfaces repo. @ros2/team any thoughts about this?

That's fine by me.

mikaelarguedas:
closing this in favor of https://github.com/ros2/rcl_interfaces/pull/16

PR #31, opened 2017-04-17
Add distortion_models.hpp to sensor_msgs package.

clalancette:
This is a place to put utilities for sensor_msgs into one place, while leaving sensor_msgs "pure" message definitions.

Note that right now I just added some distortion model stuff, necessary to move ros2/vision_opencv#1 moving along.  We'll be adding more to this soon.

connects to ros2/ros2#342


mikaelarguedas:
can this match the name of the packages it's referring too ? `sensor_msgs` rather than `sensor_msg`

tfoote:
Where are these defined in the ROS1 package? We should try to mirror that structure as much as possible.

mikaelarguedas:
> Where are these defined in the ROS1 package? We should try to mirror that structure as much as possible.

I tend to agree with this. The argument that was made against it is that people not using the c++ implementation of the messages don't want to have to compile C++ utilities when generating their messages.

clalancette:
@mikaelarguedas About the name of the package; I did "sensor_msg_utils" since that seemed to flow a bit better, but I can definitely change it to "sensor_msgs_utils" if we think that is better (it will help with tab completion, too).

@tfoote In ROS1, these definitions are under sensor_msgs/include/...  Currently in ROS2, all of the packages in common_interfaces are strictly .msg or .srv files.  As Mikael said, I wasn't sure if we wanted to match the current ROS1 structure, or if we wanted to keep the _msgs directories pure for code-generating purposes.  I went with the latter, though I don't have strong opinions on it either way.

dirk-thomas:
I assume that nobody ported the additional sources in the first place. But there is no reason a message package can't contain additional code (C++, Python, whatever else). That is already the case in ROS 1.

mikaelarguedas:
> About the name of the package; I did "sensor_msg_utils" since that seemed to flow a bit better,

 I think that it'll just be more confusing if someone then create a sensor_msg package or message. I think being consistent is valuable in this case.

This is just a remark given that it looks like this will end up (based on the comments) being in the `sensor_msgs` package itself as it is in ROS1.

clalancette:
Yeah, that's fine.  I have no problem getting rid of this and just putting it into sensor_msgs instead.  I'll do that.

mikaelarguedas:
@clalancette I'm putting this back in progress then

clalancette:
Yep, thanks.

clalancette:
The build I'm doing right now for image_geometry should actually fill in nicely as CI for this as well, since it also requires this include file:

CI:
windows: [Build Status](http://ci.ros2.org/job/ci_windows/2620/)
linux: [Build Status](http://ci.ros2.org/job/ci_linux/2506/)
linux aarch64: [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/101/)
osx: [Build Status](http://ci.ros2.org/job/ci_osx/1974/)

clalancette:
I know @wjwwood already approved, but that was before the latest changes.  Another review and ACK would be appreciated.  Thanks!

clalancette:
Thanks, merging :).

dirk-thomas:
While being late to the game:
* Why should the constants be changed from `const char[]`? `std::string` seems to be problematic too regarding initialization time?
* If the type is not changed the copyright shouldn't be appended since there is no change except whitespace changes.

clalancette:
The only reason I changed it to const char[] is because one of the linters (I forget which one) complained that we should use const char[], not std::string, for constant strings.  I actually don't care either way, though if we go back to std::string, we'll have to quiet the linter.

And yeah, I agree with you, if we don't change the type, we can get rid of the extra copyright, since we made no real changes.

clalancette:
Oh, here it is.  I guess it was cpplint:

25: Test command: /usr/bin/python3 "-u" "/home/clalancette/ros2_ws/install/share/ament_cmake_test/cmake/run_test.py" "/home/clalancette/ros2_ws/build/sensor_msgs/test_results/sensor_msgs/cpplint.xunit.xml" "--output-file" "/home/clalancette/ros2_ws/build/sensor_msgs/ament_cpplint/cpplint.txt" "--command" "/home/clalancette/ros2_ws/install/bin/ament_cpplint" "--xunit-file" "/home/clalancette/ros2_ws/build/sensor_msgs/test_results/sensor_msgs/cpplint.xunit.xml"
25: Test timeout computed to be: 60
25: -- run_test.py: invoking following command in '/home/clalancette/ros2_ws/src/ros2/common_interfaces/sensor_msgs':
25:  - /home/clalancette/ros2_ws/install/bin/ament_cpplint --xunit-file /home/clalancette/ros2_ws/build/sensor_msgs/test_results/sensor_msgs/cpplint.xunit.xml
25: /home/clalancette/ros2_ws/src/ros2/common_interfaces/sensor_msgs/include/sensor_msgs/distortion_models.hpp:43:  For a static/global string constant, use a C style string instead: "const char PLUMB_BOB[]".  [runtime/string] [4]
25: /home/clalancette/ros2_ws/src/ros2/common_interfaces/sensor_msgs/include/sensor_msgs/distortion_models.hpp:44:  For a static/global string constant, use a C style string instead: "const char RATIONAL_POLYNOMIAL[]".  [runtime/string] [4]
25: Category 'runtime/string' errors found: 2
25: Total errors found: 2
25: Using '--root=sensor_msgs/include' argument
25: 
25: Done processing /home/clalancette/ros2_ws/src/ros2/common_interfaces/sensor_msgs/include/sensor_msgs/distortion_models.hpp
25: 
25: -- run_test.py: return code 1
25: -- run_test.py: verify result file '/home/clalancette/ros2_ws/build/sensor_msgs/test_results/sensor_msgs/cpplint.xunit.xml'
25/28 Test #25: cpplint ...........................................***Failed    0.26 sec

dirk-thomas:
Should that be filled and resolved upstream?

clalancette:
I'm honestly not sure.  Is there any advantage to const char[] over std::string for a global/static string?

dirk-thomas:
Because of "static initialization order fiasco" `const char[]` seems to be the better choice.

Issue #35, opened 2017-05-11
Re-enable copyright linting in sensor_msgs


clalancette:
Now that ament/ament_lint#73 has been merged, we should be able to re-enable copyright linting in the sensor_msgs package.  It is probably just a one line change.

dirk-thomas:
Please create a PR for this change which just needs to remove [these two lines](https://github.com/ros2/common_interfaces/blob/96a6be9ee65f27605af24fe3b3a172697534c78a/sensor_msgs/CMakeLists.txt#L61-L62).

PR #36, opened 2017-05-16
[sensor_msgs] Reenable copyright test

mikaelarguedas:
addresses https://github.com/ament/ament_lint/pull/73#issuecomment-300580964
Fixes https://github.com/ros2/common_interfaces/issues/35

### DIFF
--- a/sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.hpp
+++ b/sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.hpp
@@ -13,7 +13,7 @@
 //    copyright notice, this list of conditions and the following
 //    disclaimer in the documentation and/or other materials provided
 //    with the distribution.
-//  * Neither the name of the Willow Garage nor the names of its
+//  * Neither the name of the Open Source Robotics Foundation, Inc. nor the names of its
 //    contributors may be used to endorse or promote products derived
 //    from this software without specific prior written permission.
 //
--- a/sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.hpp
+++ b/sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.hpp
@@ -1,36 +1,34 @@
-/*
- * Software License Agreement (BSD License)
- *
- *  Copyright (c) 2013, Open Source Robotics Foundation
- *  All rights reserved.
- *
- *  Redistribution and use in source and binary forms, with or without
- *  modification, are permitted provided that the following conditions
- *  are met:
- *
- *   * Redistributions of source code must retain the above copyright
- *     notice, this list of conditions and the following disclaimer.
- *   * Redistributions in binary form must reproduce the above
- *     copyright notice, this list of conditions and the following
- *     disclaimer in the documentation and/or other materials provided
- *     with the distribution.
- *   * Neither the name of Open Source Robotics Foundation nor the names of its
- *     contributors may be used to endorse or promote products derived
- *     from this software without specific prior written permission.
- *
- *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
- *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
- *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
- *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
- *  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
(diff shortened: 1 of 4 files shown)


sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.hpp
```diff
@@ -1,36 +1,34 @@
...
+// modification, are permitted provided that the following conditions
+// are met:
+//
+//  * Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+//  * Redistributions in binary form must reproduce the above
+//    copyright notice, this list of conditions and the following
+//    disclaimer in the documentation and/or other materials provided
+//    with the distribution.
+//  * Neither the name of the Willow Garage nor the names of its
```
clalancette: Cut-n-paste error; the copyright holder here should be OSRF, not Willow Garage.

sensor_msgs/include/sensor_msgs/point_cloud2_iterator.hpp
```diff
@@ -1,36 +1,34 @@
...
+// are met:
+//
+//  * Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+//  * Redistributions in binary form must reproduce the above
+//    copyright notice, this list of conditions and the following
+//    disclaimer in the documentation and/or other materials provided
+//    with the distribution.
+//  * Neither the name of the Willow Garage nor the names of its
+//    contributors may be used to endorse or promote products derived
```
clalancette: Same here.

sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.hpp
mikaelarguedas: fixed in https://github.com/ros2/common_interfaces/pull/36/commits/63541aae7451a9834170a6d54c8bdb525aa179a3. I didn't match the 80-character line length because the script is not robust to new lines within the licence

dirk-thomas:
The headers were copied as-is from upstream (e.g. https://github.com/ros/common_msgs/blob/90063ba9962274e809f91a4045f2d09dd72833c3/sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.h). Therefore the copyright should stay as-is. OSRF has done nothing which justifies adding copyright and especially can't remove the previous copyright owner. Please update the headers to restore the original copyright from upstream.

sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.hpp
clalancette: Actually, making ament_lint robust to newlines should be fairly straightforward.  I've opened an issue for that here: ament/ament_lint#75

clalancette:
@dirk-thomas The original copyright was OSRF; I think @mikaelarguedas just did a copy-n-paste, which introduced Willow Garage.  I just had him switch it back to the original.

mikaelarguedas:
> @dirk-thomas The original copyright was OSRF; I think @mikaelarguedas just did a copy-n-paste, which introduced Willow Garage. I just had him switch it back to the original.

Exactly, I was tired of trying to make the test pass so I copied the one I had working and forgot to replaced the second occurence of the copyright holder as @clalancette pointed out in his review

sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.hpp
mikaelarguedas: Yes it should be fairly easy. My goal here was to stop adding code with copyrights from elsewhere without reenabling the tests, not improving the copyright test itself.

dirk-thomas:
Where is e.g. https://github.com/ros2/common_interfaces/blob/fdadc05b0987c2069426ff4e101680b05312a525/sensor_msgs/include/sensor_msgs/distortion_models.hpp#L2 coming from?

dirk-thomas:
I guess it is just really difficult to see in the history. The first imported version of the file is already modified. When importing files like this from another source please keep the unmodified import and any changes in separate commits. Otherwise it is unnecessary more complicated to trace changes.

clalancette:
Exactly the same place: https://github.com/ros/common_msgs/blob/90063ba9962274e809f91a4045f2d09dd72833c3/sensor_msgs/include/sensor_msgs/distortion_models.h

That one is highly debatable whether we should add the copyright for OSRF or not, but since the file is so simple and we replaced basically all of the "meat", I left in copyright OSRF along with the Willow Garage one.  We could remove that.  Also, I forgot to add a comment pointing to where that file originally came from (as you mentioned); I'll do that as well.

I'm not totally opposed to importing the original and then modifying it (as you say, it makes the history easier), but it does break the bisectability of the repository, which is why I generally haven't been doing it.

mikaelarguedas:
agreed, the file changed location, extension and content so history is completely untraceable.

clalancette:
ros2/common_interfaces#37

PR #37, opened 2017-05-16
Add a comment saying where distortion_model.hpp originally came from.

clalancette:
Signed-off-by: Chris Lalancette <clalancette@osrfoundation.org>

### DIFF
--- a/sensor_msgs/include/sensor_msgs/distortion_models.hpp
+++ b/sensor_msgs/include/sensor_msgs/distortion_models.hpp
@@ -31,7 +31,7 @@
 // ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
-// This file is originally ported from ROS1:
+// This file is originally from:
 // https://github.com/ros/common_msgs/blob/jade-devel/sensor_msgs/include/sensor_msgs/distortion_models.h
 
 #ifndef SENSOR_MSGS__DISTORTION_MODELS_HPP_
--- a/sensor_msgs/include/sensor_msgs/distortion_models.hpp
+++ b/sensor_msgs/include/sensor_msgs/distortion_models.hpp
@@ -32,7 +32,7 @@
 // POSSIBILITY OF SUCH DAMAGE.
 
 // This file is originally from:
-// https://github.com/ros/common_msgs/blob/jade-devel/sensor_msgs/include/sensor_msgs/distortion_models.h
+// https://github.com/ros/common_msgs/blob/7dea912/sensor_msgs/include/sensor_msgs/distortion_models.h
 
 #ifndef SENSOR_MSGS__DISTORTION_MODELS_HPP_
 #define SENSOR_MSGS__DISTORTION_MODELS_HPP_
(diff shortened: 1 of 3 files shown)


clalancette:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2602/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/183/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2060/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2719/)

sensor_msgs/include/sensor_msgs/distortion_models.hpp
```diff
@@ -31,6 +31,9 @@
 // ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+// This file is originally ported from ROS1:
```
mikaelarguedas: not sure "ported" applies here, there is nothing ROS (1 or 2) specific in this file

clalancette: True, but the original file is from a ROS1 repository.  Should I change it to just:

"This file is originally from:"

?

mikaelarguedas: yes or "this file was originally copied from". Not sure about the exact wording but I think a more generic phrasing will allow for more reuse throughout the codebase

clalancette:
I changed the wording in all three places it was used before.  New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2604/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/184/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2061/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2720/)

dirk-thomas:
Do we need CI builds for a comment-only change?

clalancette:
Heh, the last time I didn't do one, I ended up screwing things up (the linters can still fire off for comment-only changes).  So I got conservative and ended up doing it for everything.

mikaelarguedas:
I think it's good to make sure not to break the tests by modifying comment sections. Then I agree that running on all platforms and all tests is overkill. A single platform (Windows is usually the pickiest) testing only the modified package would be more sensible IMO.

sensor_msgs/include/sensor_msgs/distortion_models.hpp
```diff
@@ -31,6 +31,9 @@
 // ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+// This file is originally from:
+// https://github.com/ros/common_msgs/blob/jade-devel/sensor_msgs/include/sensor_msgs/distortion_models.h
```
dirk-thomas: Please include the exact hash when this file was forked.

clalancette:
I ended up encoding the hashes into the URLs themselves.  That way we know exactly from whence it came.

PR #39, opened 2017-06-12
added uuid_msgs

BrannonKing:
for GUID/UUID support

### DIFF
--- a/uuid_msgs/CMakeLists.txt
+++ b/uuid_msgs/CMakeLists.txt
@@ -1,0 +1,24 @@
+cmake_minimum_required(VERSION 3.5)
+
+project(uuid_msgs)
+
+if(NOT WIN32)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14 -Wall -Wextra -Wpedantic")
+endif()
+
+find_package(ament_cmake REQUIRED)
+find_package(builtin_interfaces REQUIRED)
+find_package(rosidl_default_generators REQUIRED)
+
+set(msg_files
+  "msg/UniqueID.msg"
+)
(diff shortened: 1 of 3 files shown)


uuid_msgs/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+cmake_minimum_required(VERSION 3.5)
+
+project(uuid_msgs)
+
+if(NOT WIN32)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14 -Wall -Wextra -Wpedantic")
+endif()
+
+find_package(ament_cmake REQUIRED)
+find_package(builtin_interfaces REQUIRED)
```
mikaelarguedas: I think that this package doesn't need to depend on `builtin_interfaces` because it's using only primitive types, same below and in the package.xml

tfoote:
We generally don't want to diverge ROS1 and ROS2 structures. Especially as they go for maintenance etc. Looking toward the future it's likely that we'll want to have ROS1 and ROS2 message definitions in the same repos on different branches to keep them in sync. Thus I would suggest that we not just pull this package into this repo unless we plan to do the same in ROS1.

dirk-thomas:
I don't think we should pull that package into this repo. Instead the upstream repo containing the ROS 1 should be forked.

BrannonKing:
Okay then. Let's fork the other repo.

tfoote:
I've forked it to here: https://github.com/ros2/unique_identifier if you could resubmit the above changes as a PR on the new ros2 branch that would be great.

BrannonKing:
Pull request is in: https://github.com/ros2/unique_identifier/pull/1

PR #40, opened 2017-08-25
Make all of the messages and services have a runtime dependency on ro…

clalancette:
…sidl_runtime_cpp.

Going forward, they are all going to need a header file
from rosidl_runtime_cpp.  Therefore, we make sure they all
properly export this as a runtime dependency.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

connects to ros2/ros2#393

See CI in ros2/rosidl#231


clalancette:
Since we decided to delay this until later, closing this PR; see ros2/ros2#393 for future work.

PR #41, opened 2017-08-31
create common_interfaces metapackage

mikaelarguedas:
this came up a few times and more often since we started shipping binaries.
Creating this metapackage has 2 benefits:
- allow people to install all common message packages with a single debian package
- show an example of how metapackages look like in ROS2

One remaining question though: we renamed all message packages from `foo_interfaces` to `foo_msgs` a while ago to simplify the mapping between ros1 and ros2 messages. Should this package be named common_msgs like its' ROS1 equivalent? If so should we also rename this repository accordingly ?

### DIFF
--- a/common_interfaces/CMakeLists.txt
+++ b/common_interfaces/CMakeLists.txt
@@ -1,0 +1,7 @@
+cmake_minimum_required(VERSION 3.5)
+
+project(common_interfaces)
+
+find_package(ament_cmake REQUIRED)
+
+ament_package()
--- a/common_interfaces/package.xml
+++ b/common_interfaces/package.xml
@@ -1,0 +1,30 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>common_interfaces</name>
+  <version>0.0.2</version>
+  <description>A package containing some message definitions used in the implementation or actions.</description>
+  <maintainer email="william@osrfoundation.org">William Woodall</maintainer>
+  <license>Apache License 2.0</license>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <exec_depend>actionlib_msgs</exec_depend>
+  <exec_depend>builtin_interfaces</exec_depend>
+  <exec_depend>diagnostic_msgs</exec_depend>
+  <exec_depend>geometry_msgs</exec_depend>
+  <exec_depend>nav_msgs</exec_depend>
+  <exec_depend>sensor_msgs</exec_depend>
+  <exec_depend>shape_msgs</exec_depend>
+  <exec_depend>std_msgs</exec_depend>
+  <exec_depend>std_srvs</exec_depend>
+  <exec_depend>stereo_msgs</exec_depend>
+  <exec_depend>trajectory_msgs</exec_depend>
+  <exec_depend>visualization_msgs</exec_depend>
+
+  <test_depend>ament_lint_common</test_depend>
+
+  <export>
+    <build_type>ament_cmake</build_type>
+  </export>
+</package>
--- a/common_interfaces/package.xml
+++ b/common_interfaces/package.xml
@@ -2,7 +2,7 @@
 <?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="2">
   <name>common_interfaces</name>
-  <version>0.0.2</version>
+  <version>0.0.3</version>
   <description>common_interfaces contains messages and services that are widely used by other ROS packages.</description>
   <maintainer email="mikael@osrfoundation.org">Mikael Arguedas</maintainer>
   <license>Apache License 2.0</license>
--- a/common_interfaces/package.xml
+++ b/common_interfaces/package.xml
@@ -4,7 +4,7 @@
   <name>common_interfaces</name>
   <version>0.0.2</version>
   <description>A package containing some message definitions used in the implementation or actions.</description>
-  <maintainer email="william@osrfoundation.org">William Woodall</maintainer>
+  <maintainer email="mikael@osrfoundation.org">Mikael Arguedas</maintainer>
   <license>Apache License 2.0</license>
 
   <buildtool_depend>ament_cmake</buildtool_depend>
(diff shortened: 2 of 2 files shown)


wjwwood:
I actually think we should avoid metapackages all together if we can. If you think about what common_msgs represents it is never something you want to _depend_ on. It's only useful (imo) for indicating a grouping of packages or by providing a convenient apt-get target package to install. There's no need to ever actually build it, and really no need for it to have a cmake file (we restrict what it can contain anyways).

I'd rather see us address those use cases without metapackages which can never be used directly in code and make the concept of a "package" unnecessarily more complicated to explain.

For documentation purposes, i.e. grouping of packages, like the package clouds on the ROS 1 wiki, I'd rather see a tagging system or something like that.

For apt-get able packages, perhaps we should just bring back the concept of variants and have the build farm generate those directly.

For people pulling from source, the tools like rosinstall generator just need to be able to understand these variants and pull the packages that make them up.

Another alternative is to just make common_msgs a normal package, but again I think that just leads to useless packages that I wouldn't recommend people use because it encourages them to depend on more than they need.

gbiggs:
I agree  with @wjwwood. Such a metapackage would encourage people to depend on more than they actually use, which makes maintenance harder.

wjwwood:
A little bit of history on metapackages, just for context:

We made them to have a representation for stacks from the rosbuild days. Eventually we converted the variants (as they used to be called), e.g. robot, desktop, viz, simulation, etc..., to metapackages as well.

But from my perspective it was because we wanted to consolidate the concept of variants and metapckages (old stacks) but we could not remove the existing metapackages like common_msgs. So perhaps this is a chance to go the other way, have just "packages" and have another high level concept which the build farm and rosinstall generator understand (maybe a yaml file or something in rosdistro) which is used to group packages in the wiki and used to give convenient targets for apt-get and rosinstall generator.

gbiggs:
It may help to clarify the two (I think) relevant use cases:

1) Depending on a package that provides something you need. In this case, it is best to depend on as small a unit as possible, so depending on something like an "all commonly used messages" package would be bad.

2) Installing a bunch of related packages so you can use their collective functionality, such as a navigation stack implementation. In this case, ease of use suggests having a single entity that you can apt-get to bring in the lot.

mikaelarguedas:
This resulted from an offline conversation that I thought (wrongly) was to specific to be integrated to a high-level migration discussion. What prompted me to start this was the appearance of questions or hacks to [install set of packages](https://github.com/osrf/docker_images/pull/68/files#diff-d3c0a2ef6b4a2afbae88e83169cbbad3R45) as well a people porting ROS 1 repositories by simply deleting the metapackages without providing an alternative.

I agree that we don't expect/want people to depend on metapackages. And that the main use case for `metapackages` is the ability to apt-get a single entity and bring in a group of things you want to install.
I can see only very few corner cases, if any, where depending on a metapackage would make sense, and I don't think such use cases do justify the overhead of creating ROS 2 metapackages and the associated CMake boilerplate.

Variants or any alternative not relying on creating actual packages and allowing the infrastructure and tools to "group" a set of packages together will satisfy the requirement. I am not familiar with how variants were defined during the rosbuild days and am happy to hear about how it was done or how you guys think it could be done better in ROS 2.

wjwwood:
Some other info about how variants used to work:

- http://www.ros.org/reps/rep-0108.html
- http://www.ros.org/reps/rep-0131.html

Basically they were defined in yaml and there was web service called `gen_rosisntall` that could convert those variants (as well as individual packages) to rosinstall files, pretty similar to what `rosinstall_generator` can do now.

wjwwood:
I couldn't find the original repository for `gen_rosinstall`, but here's a fork:

https://github.com/mbeards/gen_rosinstall

And some other mentions:

- https://answers.ros.org/question/9672/gen_rosinstallpy-generating-invalid-install-files/
- https://answers.ros.org/question/82421/fuerte-source-install-gen_rosinstallpy-not-found/

mikaelarguedas:
thanks @wjwwood for the pointers!

tfoote:
A way to provide grouping is valuable for installation both for binaries as well as the rosinstall generator level.

One other thing mentioned is that when building from source grouping can also be valuable to do something like `--end-with group_name` is also useful when compiling for a demo etc.

Often it's nice to have metapackages in a repo with other packages. But it can get tricky if the metapackage wants to bring in a package from another repo. If that package wants to depend on a low level package in the main repo the release process deadlocks. 

`metapackage` depends on `foo` and `bar`
`foo` is in the same repo
`bar` is in another repo
`bar` depends on `foo`

`bar` is unreleaseable until `foo` is released and `metapackage` is unreleasable until `bar` is released. 

However having giving that limitation I think I'd still rather see the `metapackage` concept continue as a released package instead of having it added to the rosdistro explicitly. This keeps the dependencies and versioning inside the same git repo with the coordinated changelog browsing. In this way they can also track the behavior across ROS distros clearly and live in the same branch/version structure as the packages themselves.

For our core `variants` this is less of an argument they're well specified and known ahead of time. But for more leaf packages they're much more dyamic and having 2 ways to add dependencies would be odd. Also if we centralized it we'd need a new release process with versioning etc for them as well.

dirk-thomas:
> `metapackage` depends on `foo` and `bar`
> `foo` is in the same repo
> `bar` is in another repo
> `bar` depends on `foo`
> 
> `bar` is unreleaseable until `foo` is released and `metapackage` is unreleasable until `bar` is released.

While the packages don't have a circular dependency the repositories have a circular dependency. This problem is not related to metapackages but exists the same way with normal packages.

> However having giving that limitation I think I'd still rather see the `metapackage` concept continue as a released package instead of having it added to the rosdistro explicitly. This keeps the dependencies and versioning inside the same git repo with the coordinated changelog browsing. In this way they can also track the behavior across ROS distros clearly and live in the same branch/version structure as the packages themselves.

:+1:

wjwwood:
I was more concerned with the conceptual overhead of having a "special" kind of package which complicates it when you need to document, explain, or interact with them, rather than the effort required to make them in the first place. I'm more favorable of having them become normal packages because that removes the need for yet more special case code we have in all of our tools to handle metapackages differently, though I still feel they would be dead weight as normal packages too.

I still think the simplicity of using variants defined in the rosdistro outweighs the value of being able to do `--end-with common_msgs`. That use case could also be addressed if we implemented the groups, then you could build all of your local message packages with `--end-with-group interface_packages` or something similar.

I guess metapackages always seemed like a hack to me which was required to support migration from stacks, but if the group would rather keep them, that's fine too.

mikaelarguedas:
closing in favor of https://github.com/ros2/ros2/issues/408

mikaelarguedas:
reopening this in light of decision made in ros2/ros2#408

PR #43, opened 2017-10-26
Control shared/static linking via BUILD_SHARED_LIBS

esteve:
Not ready for review, just submitted this so that it's added to ros2/ros2#306

Connects to ros2/ros2#306


esteve:
Not actually needed, https://github.com/ros2/rosidl_typesupport/pull/14/commits/ecb19314b7db2cabad9c87d2491d283f082504af should handle this for all messages transparently.

PR #50, opened 2018-03-26
add rosgraph_msgs package with Clock message

dirk-thomas:
Fixes ros2/ros1_bridge#104. Connect to ros2/ros1_bridge#104.

This adds the `Clock` message from the [rosgraph_msgs](https://github.com/ros/ros_comm_msgs/blob/indigo-devel/rosgraph_msgs) package.

The other two messages fro ROS 1 seem to not have a 1-to-1 equivalent in ROS 2 therefore I didn't add them at this point.

### DIFF
--- a/rosgraph_msgs/msg/Clock.msg
+++ b/rosgraph_msgs/msg/Clock.msg
@@ -1,4 +1,3 @@
-# roslib/Clock is used for publishing simulated time in ROS.
 # This message simply communicates the current time.
 # For more information, see http://www.ros.org/wiki/Clock
 builtin_interfaces/Time clock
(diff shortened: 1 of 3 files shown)


rosgraph_msgs/msg/Clock.msg
```diff
@@ -0,0 +1,4 @@
+# roslib/Clock is used for publishing simulated time in ROS.
```
mikaelarguedas: copy-n-paste ?

dirk-thomas: Yes, copied 1-to-1 from ROS 1.

mikaelarguedas: Please update the comment as there is no such thing as roslib/Clock in ROS 2. And this doesn't exist in ROS1 since C-Turtle either from what I can tell

dirk-thomas: Done: ae2776765040d4d7c09cb948f9b1b23e6008c75e

dirk-thomas:
> Should this be blacklisted on CI like the other messages unused by our current test suite ?

I would rather not blacklist it since it is only a single message.

mikaelarguedas:
> I would rather not blacklist it since it is only a single message.

Fine by me

PR #54, opened 2018-08-15
Include sstream on header that needs it

chapulina:
Prevents the following compile error when using the header:

~~~
/opt/ros/bouncy/include/sensor_msgs/impl/point_cloud2_iterator.hpp:73:23: error: aggregate 'std::stringstream err' has incomplete type and cannot be defined
    std::stringstream err;
                      ^~~
~~~

### DIFF
--- a/sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.hpp
+++ b/sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.hpp
@@ -38,6 +38,7 @@
 
 #include <sensor_msgs/msg/point_cloud2.hpp>
 #include <cstdarg>
+#include <sstream>
 #include <string>
 #include <vector>
 


mikaelarguedas:
Can you please target this at the master branch instead of the bouncy one ?

chapulina:
It would be nice to have this on Bouncy, assuming there are patch releases being made. Maybe it can be merged into both branches?

mikaelarguedas:
We take all contributions on the default branch (master) and then evaluate what gets backported.

As this bug prevents consumers from using this utiliy without adding includes themselves it will definitely be considered for backport.

chapulina:
Got it, thanks for the clarification.

chapulina:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5053/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1849/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4195/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5047/)

PR #58, opened 2018-09-12
Fill image

chapulina:
Port fill image header from ROS 1.

I'm using this on `gazebo_plugins/gazebo_ros_camera`.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5164/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1940/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4290/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5144/)

### DIFF
--- a/sensor_msgs/include/sensor_msgs/fill_image.hpp
+++ b/sensor_msgs/include/sensor_msgs/fill_image.hpp
@@ -32,6 +32,9 @@
 *  POSSIBILITY OF SUCH DAMAGE.
 *********************************************************************/
 
+// This file is originally ported from ROS1:
+// https://github.com/ros/common_msgs/blob/89069bc/sensor_msgs/include/sensor_msgs/fill_image.h
+
 #ifndef FILLIMAGE_HH
 #define FILLIMAGE_HH
 
--- a/sensor_msgs/include/sensor_msgs/fill_image.hpp
+++ b/sensor_msgs/include/sensor_msgs/fill_image.hpp
@@ -1,74 +1,87 @@
-/*********************************************************************
-* Software License Agreement (BSD License)
-*
-*  Copyright (c) 2008, Willow Garage, Inc.
-*  All rights reserved.
-*
-*  Redistribution and use in source and binary forms, with or without
-*  modification, are permitted provided that the following conditions
-*  are met:
-*
-*   * Redistributions of source code must retain the above copyright
-*     notice, this list of conditions and the following disclaimer.
-*   * Redistributions in binary form must reproduce the above
-*     copyright notice, this list of conditions and the following
-*     disclaimer in the documentation and/or other materials provided
-*     with the distribution.
-*   * Neither the name of the Willow Garage nor the names of its
-*     contributors may be used to endorse or promote products derived
-*     from this software without specific prior written permission.
-*
-*  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
-*  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
-*  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
-*  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
-*  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
-*  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
-*  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
-*  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
-*  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
-*  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
-*  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
-*  POSSIBILITY OF SUCH DAMAGE.
-*********************************************************************/
+// Copyright (c) 2008, Willow Garage, Inc.
+// All rights reserved.
+//
+// Software License Agreement (BSD License 2.0)
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions
+// are met:
+//
+//  * Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+//  * Redistributions in binary form must reproduce the above
+//    copyright notice, this list of conditions and the following
+//    disclaimer in the documentation and/or other materials provided
+//    with the distribution.
+//  * Neither the name of the Willow Garage nor the names of its
+//    contributors may be used to endorse or promote products derived
+//    from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+// FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+// COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+// INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+// BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+// LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
Looks great!
I know this is labeled in progress but I'll give early feedback to make it easier for when it's ready for review.

Could you please split this PR in multiple commits?
- 1 commit for copying the original file with its history
- 1 commit for adding a reference to where the file has been copied from (similar to https://github.com/ros2/message_filters/pull/6)
- 1 commit for the ROS2 port

This would make the PR easier to review as well as better trace-ability.
Thanks!

sensor_msgs/include/sensor_msgs/fill_image.hpp
```diff
@@ -0,0 +1,86 @@
...
+{
+/**
+ * @brief Fills an image message.
+ * @param image Image to be filled.
+ * @param encoding_arg Encoding type, such as sensor_msgs::image_encodings::RGB8.
+ * @param rows_arg Number of rows.
+ * @param cols_arg Number of columns.
+ * @param step_arg Step size.
+ * @param data_arg Data to fill image with.
+ * @return True if successful.
```
wjwwood: nitpick: Typically we use this style:

```c++
/// Fills an image message.
/**
 * \param image Image to be filled.
 * ...
 * \return True if successful.
 */
static inline bool fillImage(
  ...
```

Though we're not 💯percent consistent, here are some good examples (imo):

- https://github.com/ros2/rcutils/blob/f04cc666dd06347050ff5c18b80079922e729d92/include/rcutils/get_env.h#L26-L58
- https://github.com/ros2/rcl/blob/e5e338a8e0d0b4a4824edae7f351963ca2435015/rcl/include/rcl/validate_topic_name.h#L38-L88
- https://github.com/ros2/rclcpp/blob/b1af28047c26da0fe5723e3ea95ed1143bc224e5/rclcpp/include/rclcpp/node.hpp#L70-L76

I believe this is referred to as the "special C++ style comment" in the doxygen documentation (second instance of "option 3"):

https://www.stack.nl/~dimitri/doxygen/manual/docblocks.html#cppblock

It avoids the `\brief`, and for documentation that only has a brief, it avoids `/** ... */` stuff.

Also we prefer the `\style` because that's what the doxygen documentation uses, e.g.:

https://www.stack.nl/~dimitri/doxygen/manual/commands.html

I think we somewhat mention this in our developer guide, but it could be clearer:

https://github.com/ros2/ros2/wiki/Developer-Guide#comments-and-doc-comments

Mostly just to be consistent with the rest of the newly written code.

sensor_msgs/include/sensor_msgs/fill_image.hpp
```diff
@@ -0,0 +1,86 @@
...
+// This file is originally ported from ROS1:
+// https://github.com/ros/common_msgs/blob/jade-devel/sensor_msgs/include/sensor_msgs/fill_image.h
+
+#ifndef SENSOR_MSGS__FILL_IMAGE_HPP_
+#define SENSOR_MSGS__FILL_IMAGE_HPP_
+
+#include <sensor_msgs/msg/image.hpp>
+#include <sensor_msgs/image_encodings.hpp>
+
+#include <string>
```
wjwwood: I believe system includes should be listed before external third-party headers and project headers. The only exception would be if this were an implementation of a header, you could include that header first. See:

https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

wjwwood: We have a test for this in cpplint (`ament_cpplint`), but sadly it doesn't work very well.

sensor_msgs/include/sensor_msgs/fill_image.hpp
mikaelarguedas: Yeah it's a matter of keeping the file close to its' ROS1 equivalent vs adopting the ROS2 style. It looks like the other utilities in sensor_msgs package (e.g. [point_cloud2_iterator.hpp](https://github.com/ros2/common_interfaces/blob/3d9f88ccdb65bb3a087db1066b17442e36b2a8cd/sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.hpp)) kept the original documentation style (`@`), the conjugated version `Fills` instead of `Fill` etc

wjwwood: The original file for this one doesn't have documentation at all, it was added in the migration.

The one you linked to mixes style anyways :/ https://github.com/ros2/common_interfaces/blob/3d9f88ccdb65bb3a087db1066b17442e36b2a8cd/sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.hpp#L46

wjwwood: You're right though it should really be `Fill` and not `Fills` (something we have tests for in Python but not C++), but since this is new documentation it's probably worth doing.

sensor_msgs/include/sensor_msgs/fill_image.hpp
mikaelarguedas: Yeah cpplint assume anything within `<` to be a system header. The "correct" thing to do here would be:
```c++
#include <string>

#include "sensor_msgs/msg/image.hpp"
#include "sensor_msgs/image_encodings.hpp"
```

I assumed it was using `<` to stay close to original file but apparently the ROS 1 version uses double quotes as well so I think it makes sense to change it here and put the system header first

sensor_msgs/include/sensor_msgs/fill_image.hpp
mikaelarguedas: > The original file for this one doesn't have documentation at all, it was added in the migration.

Yeah it's worth doing then, I just assumed it came from upstream

chapulina:
Thanks for all the feedback!

> Could you please split this PR in multiple commits?

Done

> Tip for CI

Very useful, thanks. Here's the latest:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5166/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1941/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4291/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5145/)

> nitpick: Typically we use this style

I was just copying the style I found in other headers in this repo. Now using the correct style: 31012b4. Btw, we've been following the recommended style on `gazebo_ros_pkgs` :wink: 

> it should really be Fill and not Fills 

Done in 31012b4. Interesting, I didn't know about this, is it documented somewhere?

> system includes should be listed before external third-party headers 

> cpplint assume anything within < to be a system header

Changed in 31012b4 as suggested. In my experience so far, the include order accepted by the linters can get tricky really fast, and I noticed that putting system headers in the bottom usually makes all linters happy. See [here](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/ros2/gazebo_plugins/src/gazebo_ros_force.cpp#L15) for example, where the order goes `<*.hh>` `<*.hpp>` `<system>` and looks super weird to me.  Also see this relevant comment: https://github.com/ros2/demos/pull/80#discussion_r81660064.

sensor_msgs/include/sensor_msgs/fill_image.hpp
```diff
@@ -32,6 +32,9 @@
 *  POSSIBILITY OF SUCH DAMAGE.
 *********************************************************************/
 
+// This file is originally ported from ROS1:
+// https://github.com/ros/common_msgs/blob/jade-devel/sensor_msgs/include/sensor_msgs/fill_image.h
```
mikaelarguedas: Nit (as this file didnt change for years): Can you please use a link to the specific commit it was copied from rather than just the branch ?

https://github.com/ros/common_msgs/blob/89069bc/sensor_msgs/include/sensor_msgs/fill_image.h

chapulina: Done: 29a141d

sensor_msgs/include/sensor_msgs/fill_image.hpp
```diff
@@ -0,0 +1,86 @@
...
+  return true;
+}
+
+/// Clear an image message.
+/**
+ * \param[out]image Image to be cleared.
+ */
+static inline void clearImage(msg::Image & image)
+{
+  image.data.resize(0);
```
mikaelarguedas: (Question: not blocking) Should the other fields of the message be cleared / reinitialized to their default values ?
if yes it doesn't have to be addressed now and can be done in a follow-up, if no can you please update the docblock to reflect what part of the message are being cleared by this function?

chapulina: Kept the ROS1 behaviour and documented it 19bb03c

chapulina:
It looks like this is ready for merging. What's the protocol here, should I do more history rewriting to merge the last 3 commits into 1?

mikaelarguedas:
> What's the protocol here, should I do more history rewriting to merge the last 3 commits into 1?

Yes ideally we would end with 3 commits in this PR:
- 1 commit for copying the original file with its history
- 1 commit for adding a reference to where the file has been copied from
- 1 commit for the ROS2 port

So the rewriting we should do is:
- 29a141d006ab515d41f16ba21c49f7a07b316200 should be part of ac64728ec5961c3535169448639f68954427e209
- 19bb03c82fc898a213042cc1e012d87c78214b64 should be part of 31012b4873909d431cc966a7e46699f5d9f515a5

Then you can merge it without squashing so that we keep these 3 commits distinct in the history

chapulina:
Thanks for the tips, I *think* I didn't break anything. Feel free to merge.

wjwwood:
> > it should really be Fill and not Fills

> Done in 31012b4. Interesting, I didn't know about this, is it documented somewhere?

It's based on something from the Python documentation standard which probably bleeds into our C++ style, but I don't think it's documented in our guide...

> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

-- https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

PR #60, opened 2018-11-02
New package, unique_identifier_msgs

stonier:
There was a suggestion over at https://github.com/ros2/ros2/issues/593 to migrate `uuid_msgs` to `common_interfaces`. 

I'd also like to see it here - making use of a unique identifier became a very common thing to do once we started doing so.

### DIFF
--- a/unique_identifier_msgs/CHANGELOG.rst
+++ b/unique_identifier_msgs/CHANGELOG.rst
@@ -1,0 +1,35 @@
+Change history
+==============
+
+1.1.0 (2018-11-01)
+------------------
+* Name change: uuid_msgs -> unique_identifier_msgs
+
+1.0.4 (2014-04-30)
+------------------
+
+ * add architecture independent tag
+
+1.0.1 (2013-03-25)
+-------------------
+
+ * Hydro release update.
+
+1.0.0 (2013-03-18)
+-------------------
+
+ * Hydro release.
+ * Convert to catkin (`#1`_).
+
+0.9.0 (2013-01-03)
+------------------
+
+ * Initial release to Groovy.
+
+0.8.0 (2012-07-19)
+------------------
+
+ * Initial release to Fuerte.
+ * Provides uuid_msgs/UniqueID message.
+
+.. _`#1`: https://github.com/ros-geographic-info/unique_identifier/issues/1
--- a/unique_identifier_msgs/CMakeLists.txt
+++ b/unique_identifier_msgs/CMakeLists.txt
@@ -1,0 +1,28 @@
+cmake_minimum_required(VERSION 3.5)
+project(unique_identifier_msgs)
+
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  # we dont use add_compile_options with pedantic in message packages
+  # because the Python C extensions dont comply with it
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wpedantic")
+endif()
+
+find_package(ament_cmake REQUIRED)
+find_package(rosidl_default_generators REQUIRED)
+find_package(std_msgs REQUIRED)
+
+set(msg_files
+  "msg/UniqueID.msg"
+)
+rosidl_generate_interfaces(${PROJECT_NAME}
+  ${msg_files}
+  DEPENDENCIES std_msgs
+)
+
+ament_export_dependencies(rosidl_default_runtime)
+
+ament_package()
--- a/unique_identifier_msgs/msg/UniqueID.msg
+++ b/unique_identifier_msgs/msg/UniqueID.msg
@@ -1,0 +1,6 @@
+# A universally unique identifier (UUID).
+#
+#  http://en.wikipedia.org/wiki/Universally_unique_identifier
+#  http://tools.ietf.org/html/rfc4122.html
+
+uint8[16] uuid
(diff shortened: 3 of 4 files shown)


dirk-thomas:
What motivated the rename of the package?

The package version must match the version of the other packages in this repo.

> One could almost make an argument to have `UniqueId.msg` in `std_msgs`.

Most messages defined in `std_msgs` are not intended to be used (except `Header`, `ColorRGBA` etc.) for actual data being testing since they don't provide any semantic.

Keeping the package as it was in ROS 1 is certainly preferred if there are no very strong reasons to diverge.

dirk-thomas:
That being said the ROS 2 version of the package could just be created on a branch in the existing repository. That would also follow the recommendation for porting.

stonier:
> std_msgs

**[Off the Table]** Fine for leaving it out of this discussion.

> That being said the ROS 2 version of the package could just be created on a branch in the existing repository. That would also follow the recommendation for porting.

**[Pending]** That is already there. Before proceeding with working out transfer of maintainership for that with Jack, I'd like to explore whether we'd like to change certain things in the ROS1-ROS2 transition while we have a small window while there is no dependencies on this and people transitioning will have to make transitional changes regardless.

> What motivated the rename of the package?

**[Low Priority]** To be consistent with the language specific libraries that wrap usage of this message. As per https://github.com/ros2/ros2/issues/593, `unique_identifer_py`, `unique_identifier_cpp`, `unique_identifier_msgs` makes most sense. Reason to do so - see above.

> `unique_identifier_msgs` in a msg-only repository (here or on its own)

**[High Priority]** This is the only must-have in my opinion. Coupling message and language specific implementations in the same repository causes many development workflow problems.

> `unique_identifier_msgs` in `common_interface`

**[Low/High Priority]** It *is* a common interface, so would make sense to be here, but on that merit alone, not necessary (low priority). If however, @jacobperron and co. plan to use it for the `GoalID` of `actionlib_msgs`, then you most likely do not want `common_interfaces` to have a dependency to this and instead have it here (high priority). @jacobperron can you chime in on this?

jacobperron:
I think there is a high likelihood that we use the UUID ROS message in [rcl_action](https://github.com/ros2/rcl) via [GoalInfo.msg](https://github.com/ros2/rcl_interfaces/blob/master/action_msgs/msg/GoalInfo.msg). Therefore, I think a more appropriate place for the package might be [rcl_interfaces](https://github.com/ros2/rcl_interfaces).

> That being said the ROS 2 version of the package could just be created on a branch in the existing repository. That would also follow the recommendation for porting.

I think this would be fine as well, but it is perhaps better to do this in [ros2 org's fork](https://github.com/ros2/unique_identifier) considering it will be more closely tied to the ros2 core implementation.

> This is the only must-have in my opinion. Coupling message and language specific implementations in the same repository causes many development workflow problems.

I'm not sure I understand the issue. Maybe you could clarify?

dirk-thomas:
When used in `rcl*` putting the package in the `rcl_interfaces` repo sounds good to me.

stonier:
> When used in rcl* putting the package in the rcl_interfaces repo sounds good to me.

I'd kick back on that. There are many instances where I'd use this message type in a use case that has no direct connection to client library concepts / testing. It really is *common*.

If there really is strong objection to having it find a home in `common_interfaces` (have to admit, I'm struggling to see why), then I'd vote for keeping this as a standalone, one message repository somewhere.

> > This is the only must-have in my opinion. Coupling message and language specific implementations in the same repository causes many development workflow problems.

> I'm not sure I understand the issue. Maybe you could clarify?

We used to frequently cross-compile or build from java/android workspaces which only needed the msg artifacts built so it could communicate across the network to your robot's core cpu. In those situations, cloning in a mixed msg/c++/python repository into the workspace was at best, dragging in a long chain of dependencies, and at worst, failing because the workspace could not compile the c++ package. At this point you'd have to get your hands dirty and manually enumerate a list of all the packages you needed to explicitly ignore. Working with partners over message repositories (on a shared middleware API), but not requiring they have all the robot side sources/binaries (e.g. having our whole navigation stack just to build a client gui that sends goals to the navigation stack on the robot) was also a good thing.

The short of it - splitting your code and message repositories avoided lots of problems later on and came at almost no cost up front.

dirk-thomas:
> I'd kick back on that.

The rational for putting it into `rcl_interfaces` is:
* we don't want it to be in `common_interfaces` since nothing in `rcl` depends on it and putting `uuid_msgs` there would require `common_interfaces` to be released before `rcl`
* a separate repo is possible but that again would need to be released *before* `rcl` and there is already a repository with exactly that characteristic: `rcl_interfaces`

stonier:
So, practicality trumps organisational strangeness?

`rcl_interfaces` is definitely not where I'd expect to find `uuid_msgs`.

jacobperron:
> Is the common_interfaces->rcl_interfaces a problem? Pushing it into rcl_interfaces seems like it's just pushing that awkwardness conceptually. rcl_interfaces is not where I'd expect to find uuid_msgs

As @dirk-thomas points out, if we want to use `uuid_msgs` in `rcl`, `rclcpp`, `rclpy`, etc, then it would be better if we didn't pull in the other packages in `common_interfaces` creating dependencies that don't need to exist. Since it is expected to be used in many other contexts (outside `rcl*`) I think a good compromise would be to put `uuid_msgs` (or `unique_identifier_msgs`) in it's own repo in the ros2 org.

If this sounds like reasonable solution, I can create the repo.

stonier:
Closing this out with the understanding that @jacobperron will move forward creating a message only package of the same in the ros2 organisation.

tfoote:
Note that there's a difference between the common_interfaces repo and meta package. If it's in a standalone repository, the metapackage can include uuid_msgs so it can be promoted to be in common_interfaces w/o being in the repository. Thus for taxonomy we can have it in the core, but dependency wise for source checkouts it can be separate.

PR #61, opened 2018-11-09
add ackermann_msgs from ros1

ipa-jba:
This is a copy from https://github.com/ros-drivers/ackermann_msgs by @jack-oquin


dirk-thomas:
The package was not part of `common_msgs` and I don't see why it should be in ROS 2. Please consider to contribute the patch to a `ros2` branch in the existing repository instead.

ipa-jba:
done https://github.com/ros-drivers/ackermann_msgs/pull/7

PR #62, opened 2018-11-09
Add "SetDatatype" services

ipa-jba:
for common datatypes (Float, Double, Int, String)

Those were part of http://wiki.ros.org/cob_srvs but are imho general enough to be in "common_interfaces"


dirk-thomas:
I similar request for the ROS 1 package has not been accepted in the past: see ros/ros_comm_msgs#9

I think the same rational still applies here.

PR #63, opened 2018-11-14
update some messages to match comment extraction heuristic

dirk-thomas:
An example patch to match the heuristic to extract comments from interface files as described in ros2/rosidl#316.

### DIFF
--- a/sensor_msgs/msg/ChannelFloat32.msg
+++ b/sensor_msgs/msg/ChannelFloat32.msg
@@ -3,7 +3,7 @@
 # array should be the same as the length of the points array in the
 # PointCloud, and each value should be associated with the corresponding
 # point.
-
+#
 # Channel names in existing practice include:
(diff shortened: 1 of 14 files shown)


dirk-thomas:
The second commit updates more message / service files in the same way.

geometry_msgs/msg/Inertia.msg
```diff
@@ -1,3 +1,4 @@
+
```
jacobperron: In this case, the inertia tensor comment is only applied to the first element. Would it be better to describe the the tensor at the top of the message (ie. putting the comment block here)?

dirk-thomas: The next line contains the comment `Mass [kg]` which should be associated with the field `float64 m` rather than the message level. Therefore the new empty line.

Restructuring comments from somewhere below in the file to the message level is out of scope for this PR. Please feel free to do so in a follow up PR if it seems beneficial.

PR #70, opened 2019-07-19
Add GetInteractiveMarkers.srv

jacobperron:
This service is meant to replace use of InteractiveMarkerInit.msg on a latched topic.

Connects to https://github.com/ros-visualization/interactive_markers/issues/42

### DIFF
--- a/visualization_msgs/srv/GetInteractiveMarkers.srv
+++ b/visualization_msgs/srv/GetInteractiveMarkers.srv
@@ -1,0 +1,12 @@
+---
+# Identifier for the interactive marker server.
+string server_id
+
+# Sequence number.
+# Set to the sequence number of the latest update message
+# at the time the server received the request.
+# Clients use this to detect if any updates were missed.
+uint64 sequence_number
+
+# All interactive markers provided by the server.
+InteractiveMarker[] markers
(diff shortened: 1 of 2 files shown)


visualization_msgs/srv/GetInteractiveMarkers.srv
```diff
@@ -0,0 +1,12 @@
+---
+# Identifier for the interactive marker server.
+string server_id
```
dgossow: wouldn't this be redundant?

jacobperron: If we assume there is one server allowed per service name, then yes. But, considering it's currently possible to have multiple servers with the same name it might be good to keep the ID field to help differentiate them.

jacobperron: After settling that we will only expect one server per interactive marker "topic", this is redundant. Removed in 3388b8d

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7935/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3969/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6460/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7784/)

I'm not sure why, but the Windows build starting failing to generate visualization_msgs/msg/Marker.msg with this PR. It seemed to not like the `DELETE` constant, so I added a guard for it in https://github.com/ros2/rosidl/pull/401. Now it is happy: [Build Status](https://ci.ros2.org/job/ci_windows/7831/)

Issue #64, opened 2018-11-14
Duplicate action message definitions


nburek:
#### Feature description
There actionlb_msgs defined in this repository appear to be a duplicate of the action_msgs defined in rcl_interfaces: https://github.com/ros2/rcl_interfaces/tree/master/action_msgs

It is confusing having both of these as it isn't clear which one should be used and may cause fragmentation going forward. If possible one of these should be deleted or at the very least renamed. 

#### Implementation considerations
I assume that one of these is here for some historical reason. If there are dependency reasons that one of these cannot be deleted yet then it would be good to at least mark them as deprecated and document what should be used. If someone knows which of these should actually be kept going forward it would be nice to have that information added to this issue.

sloretz:
`action_msgs` will be used for ROS 2 actions. The only use case for `actionlib_msgs` I'm aware of is to bridge the ROS 1 action topics directly. I think it can be removed when the ROS 1 bridge supports actions.

PR #75, opened 2019-10-02
Add notice that actionlib_msgs is deprecated

jacobperron:
Related to https://github.com/ros2/common_interfaces/issues/64

If there's better mechanism for signaling the deprecation of a package, please advise.


dirk-thomas:
https://www.ros.org/reps/rep-0140.html#deprecated

jacobperron:
> https://www.ros.org/reps/rep-0140.html#deprecated

Nifty. Added deprecation tag.

PR #86, opened 2020-03-12
Aggregate Foxy Message API Review comments.

tfoote:
This PR contains the aggregated results of the API review comments solicited [in this thread](https://discourse.ros.org/t/foxy-message-api-review/13105?u=tfoote). Please use this PR to make comments suggestions or give any other feedback on the existing comments. 

I will be going over these actions to triage them for scale and prioritization on Friday the 13th. If you would like to be involved in the meeting please ping me directly.

PR #87, opened 2020-04-02
Homogenize const types with usage in action and type storage

tfoote:
This is a follow up to the [Aggregate Foxy Message API Review](https://github.com/ros2/common_interfaces/pull/86)

In both Marker and ImageMarker the constants are all uint8

https://github.com/ros2/common_interfaces/blob/519e851f6ac71262e45b6e5ca66bce653a72948d/visualization_msgs/msg/ImageMarker.msg#L1-L8

However the action and type they are used in are int32

https://github.com/ros2/common_interfaces/blob/519e851f6ac71262e45b6e5ca66bce653a72948d/visualization_msgs/msg/ImageMarker.msg#L15-L18

This proposes to homogenize them by changing the const's declared type to be consistent.

### DIFF
--- a/visualization_msgs/msg/ImageMarker.msg
+++ b/visualization_msgs/msg/ImageMarker.msg
@@ -1,11 +1,11 @@
-uint8 CIRCLE=0
-uint8 LINE_STRIP=1
-uint8 LINE_LIST=2
-uint8 POLYGON=3
-uint8 POINTS=4
+int32 CIRCLE=0
+int32 LINE_STRIP=1
+int32 LINE_LIST=2
+int32 POLYGON=3
+int32 POINTS=4
 
-uint8 ADD=0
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Was the alternative considered - to change the type of the `type` and `action` fields. 8 bit seems more than enough to encode the possible values. If both options were considered I think it would be good to mention the rationale here why one was chosen over the other.

tfoote:
That would be possible but would be changing the message format. Which would then require bag migrations and other mechanisms. Right now the consts are just being implicitly cast to the larger type when assigned into the message and consequently this is the least disruptive resolution.

tfoote:
Rebased with basic CI to test code changes

Linux [Build Status](http://ci.ros2.org/job/ci_linux/10812/)

PR #107, opened 2020-04-11
Deprecate PointCloud message and APIs using it

### LINKED ISSUE #105: Fully deprecated PointCloud, the original, and plan for the removal in the next release
This is a follwup to the [pre Foxy Message API review](https://github.com/ros2/common_interfaces/pull/86) 

After validating that there's minimal usage of the original PointCloud we should hard deprecate it and plan for removal in ROS 2 G.

https://github.com/ros2/common_interfaces/pull/86/files#r392503804

tfoote:
Fixes #105 

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10152/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5720/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8256/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10046/)

### DIFF
--- a/sensor_msgs/include/sensor_msgs/point_cloud_conversion.hpp
+++ b/sensor_msgs/include/sensor_msgs/point_cloud_conversion.hpp
@@ -42,6 +42,16 @@
 
 #include <string>
 
+#ifndef SENSOR_MSGS_SKIP_WARNING
+#  define POINT_CLOUD_DEPRECATION_MESSAGE \
+  "PointCloud is deprecated in Foxy for PointCloud2. This whole header will be removed."
+#  ifdef _MSC_VER
+#    pragma message(POINT_CLOUD_DEPRECATION_MESSAGE)
+#  else
+#    warning POINT_CLOUD_DEPRECATION_MESSAGE
+#  endif
+#endif
+
 /**
   * \brief Convert between the old (sensor_msgs::msg::PointCloud) and the new (sensor_msgs::msg::PointCloud2) format.
   * \author Radu Bogdan Rusu
@@ -71,6 +81,7 @@ static inline int getPointCloud2FieldIndex(
   * \param input the message in the sensor_msgs::msg::PointCloud format
   * \param output the resultant message in the sensor_msgs::msg::PointCloud2 format
   */
+[[deprecated("PointCloud is deprecated as of Foxy in favor of sensor_msgs/PointCloud2.")]]
 static inline bool convertPointCloudToPointCloud2(
   const sensor_msgs::msg::PointCloud & input,
   sensor_msgs::msg::PointCloud2 & output)
@@ -125,6 +136,7 @@ static inline bool convertPointCloudToPointCloud2(
   * \param input the message in the sensor_msgs::msg::PointCloud2 format
   * \param output the resultant message in the sensor_msgs::msg::PointCloud format
   */
+[[deprecated("PointCloud is deprecated as of Foxy if favor of sensor_msgs/PointCloud2.")]]
 static inline bool convertPointCloud2ToPointCloud(
   const sensor_msgs::msg::PointCloud2 & input,
   sensor_msgs::msg::PointCloud & output)
--- a/sensor_msgs/test/test_pointcloud_conversion.cpp
+++ b/sensor_msgs/test/test_pointcloud_conversion.cpp
@@ -62,7 +62,7 @@ TEST(sensor_msgs, PointCloudConversion)
 
 #ifdef _MSC_VER
 #pragma warning(push)
-#pragma warning(disable : 4996 )
+#pragma warning(disable : 4996)
 #else
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
@@ -93,7 +93,7 @@ TEST(sensor_msgs, PointCloudConversion)
   sensor_msgs::msg::PointCloud cloud3;
 #ifdef _MSC_VER
 #pragma warning(push)
-#pragma warning(disable : 4996 )
+#pragma warning(disable : 4996)
 #else
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
--- a/sensor_msgs/test/test_pointcloud_conversion.cpp
+++ b/sensor_msgs/test/test_pointcloud_conversion.cpp
@@ -18,7 +18,17 @@
 #include "sensor_msgs/msg/point_cloud.hpp"
 #include "sensor_msgs/msg/point_cloud2.hpp"
 
+#define SENSOR_MSGS_SKIP_WARNING
+
+// #warning suppression
+// Not working due to preprocessor ignoring pragmas in g++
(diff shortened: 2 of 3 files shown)


sensor_msgs/include/sensor_msgs/point_cloud_conversion.hpp
```diff
@@ -71,6 +75,7 @@ static inline int getPointCloud2FieldIndex(
   * \param input the message in the sensor_msgs::msg::PointCloud format
   * \param output the resultant message in the sensor_msgs::msg::PointCloud2 format
   */
+RCUTILS_DEPRECATED_WITH_MSG("PointCloud is deprecated as of Foxy to be removed after. Please use sensor_msgs/PointCloud2.")
```
ivanpauno, line 78: C++14 has already a `deprecated` attribute: https://en.cppreference.com/w/cpp/language/attributes/deprecated.

Except the code here requires pre-C++14 compatibility, I would limits the usage of the new macros to pure `C` code.

tfoote:
Package only builds:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10254/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5793/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8336/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10138/)

For good measure here's a sanity check with a full downstream rebuild.
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10255/)

sensor_msgs/package.xml
```diff
@@ -15,6 +15,7 @@
   <build_depend>geometry_msgs</build_depend>
   <build_depend>std_msgs</build_depend>
 
+
```
ivanpauno, line 18: nit: seems like a leftover of a previous change

sensor_msgs/test/test_pointcloud_conversion.cpp
```diff
@@ -18,7 +18,26 @@
...
 #include "sensor_msgs/msg/point_cloud2.hpp"
 
+#ifdef _MSC_VER
+#pragma warning(push)
+#else
+#pragma GCC diagnostic push
+#ifndef __clang__
+#pragma GCC diagnostic ignored "-Wcpp"
+#endif
+#define RCUTILS_SKIP_WARNING
```
ivanpauno, line 28: is this actually a macro defined in `rcutils`?

If not, I would use a name more specific to this package.

sensor_msgs/test/test_pointcloud_conversion.cpp
```diff
@@ -18,7 +18,26 @@
 #include "sensor_msgs/msg/point_cloud.hpp"
 #include "sensor_msgs/msg/point_cloud2.hpp"
 
+#ifdef _MSC_VER
+#pragma warning(push)
```
ivanpauno, line 22: this seem to be pushing the "warning configuration" state, but not  adding any `#pragma warning( disable : ...)` in the middle

sensor_msgs/test/test_pointcloud_conversion.cpp
```diff
@@ -18,7 +18,26 @@
...
+#else
+#pragma GCC diagnostic push
+#ifndef __clang__
+#pragma GCC diagnostic ignored "-Wcpp"
```
ivanpauno, line 26: I couldn't find the docs, is this for ignoring a `#pragma message`?

ivanpauno: If the answer is yes, you can ignore this other comment https://github.com/ros2/common_interfaces/pull/107#discussion_r409746682.

sensor_msgs/include/sensor_msgs/point_cloud_conversion.hpp
```diff
@@ -36,12 +36,23 @@
...
 #include <sensor_msgs/point_field_conversion.hpp>
 
 #include <string>
 
+#define POINT_CLOUD_DEPRECATION_MESSAGE \
+  "PointCloud is deprecated in Foxy for PointCloud2. This whole header will be removed."
+#ifdef _MSC_VER
+#pragma message(POINT_CLOUD_DEPRECATION_MESSAGE)
+#else
+#ifndef RCUTILS_SKIP_WARNING
```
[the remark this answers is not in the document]
tfoote: This is specifically to the bug in g++ that the preprocessor doesn't respect the pragmas. So it's better to have the pragma suppress the warning on Windows instead of compiling it out.

 https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53431

sensor_msgs/package.xml
tfoote: yeah, i went through a lot of iterations. I'll squash this out.

sensor_msgs/test/test_pointcloud_conversion.cpp
tfoote: Yeah, I was originally going to try to prototype it here and then move it upstream, but it go so complicated I'm going to need to defer the upstream proposal until after the freeze. I'll rename it for clarity for now. And we can replace it with a potential macro.

sensor_msgs/test/test_pointcloud_conversion.cpp
tfoote: yes, this is how to ignore the `#warning` in gcc, but apparently clang doesn't support this specific diagnostic. That's probably due to them having the same problem as g++ not being able to suppress it, but they at least don't silently fail to suppress it.

tfoote:
Squashed with the feedback. Simplifying it for this specific use case and avoiding the more generic approaches that don't work in many cases.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10262/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5800/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8343/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10147/)

sensor_msgs/test/test_pointcloud_conversion.cpp
```diff
@@ -49,7 +59,20 @@ TEST(sensor_msgs, PointCloudConversion)
 
   // Convert to PointCloud2
   sensor_msgs::msg::PointCloud2 cloud2;
+
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable : 4996 )
```
ivanpauno, line 65: nit: seems to be an extra space before the closing bracket.

sensor_msgs/test/test_pointcloud_conversion.cpp
```diff
@@ -68,7 +91,19 @@ TEST(sensor_msgs, PointCloudConversion)
 
   // Convert back to PointCloud
   sensor_msgs::msg::PointCloud cloud3;
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable : 4996 )
```
ivanpauno, line 96: same

PR #113, opened 2020-05-18
Improve documentation for TransformStamped.msg

### LINKED ISSUE #94: Add documentation to TransformStamped as to why it has the extra field
This is a follwup to the [pre Foxy Message API review](https://github.com/ros2/common_interfaces/pull/86) 

https://github.com/ros2/common_interfaces/pull/86/files#r392491840

tfoote:
Fixes #94

### DIFF
--- a/geometry_msgs/msg/TransformStamped.msg
+++ b/geometry_msgs/msg/TransformStamped.msg
@@ -1,9 +1,13 @@
 # This expresses a transform from coordinate frame header.frame_id
-# to the coordinate frame child_frame_id
+# to the coordinate frame child_frame_id at the time of header.stamp
 #
 # This message is mostly used by the
-# <a href="http://www.ros.org/wiki/tf">tf</a> package.
+# <a href="http://www.ros.org/wiki/tf2">tf2</a> package.
 # See its documentation for more information.
+#
+# The child_frame_id is necessary in addition to the frame_id
+# in the Header to communication the full reference for the transform
+# in a self contained message. 
 
 std_msgs/Header header
 string child_frame_id # the frame id of the child frame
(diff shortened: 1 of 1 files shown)


geometry_msgs/msg/TransformStamped.msg
```diff
@@ -1,9 +1,13 @@
...
-# to the coordinate frame child_frame_id
+# to the coordinate frame child_frame_id at the time of header.stamp
 #
 # This message is mostly used by the
-# <a href="http://www.ros.org/wiki/tf">tf</a> package.
+# <a href="http://www.ros.org/wiki/tf2">tf2</a> package.
 # See its documentation for more information.
+#
+# The child_frame_id is necessary in addition to the frame_id
+# in the Header to communication the full reference for the transform
```
hidmic, line 9: @tfoote 
```suggestion
# in the Header to communicate the full reference for the transform
```

hidmic: I do wonder if this paragraph isn't just redundant with the definition of transform and the first paragraph.

geometry_msgs/msg/TransformStamped.msg
```diff
@@ -1,9 +1,13 @@
 # This expresses a transform from coordinate frame header.frame_id
-# to the coordinate frame child_frame_id
+# to the coordinate frame child_frame_id at the time of header.stamp
 #
 # This message is mostly used by the
-# <a href="http://www.ros.org/wiki/tf">tf</a> package.
+# <a href="http://www.ros.org/wiki/tf2">tf2</a> package.
```
hidmic, line 5: @tfoote hmm are we reusing the ROS wiki for ROS 2?

tfoote: Good point updated to use index.ros.org

geometry_msgs/msg/TransformStamped.msg
tfoote: It is somewhat redundant but people asked for clarification in the API review so I figured another sentence could help.

Issue #89, opened 2020-04-02
Move example messages out of std_msgs


tfoote:
This is a follow up to the [Aggregate Foxy Message API Review](https://github.com/ros2/common_interfaces/pull/86)

Followup to: https://github.com/ros2/common_interfaces/pull/86/files#r392505128

We should look at moving any of the messages that are not actually standard to something more like `example_msgs` this will set expectations more clearly. 

This could have wide ranging effects and it might be worth adding explicit warnings into the generators to help with the migration process.

tfoote:
Done in a combination of #116 and https://github.com/ros2/example_interfaces/pull/11

ros-discourse:
This issue has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/soss-a-whole-new-approach-to-your-ros-1-ros-2-bridge/17040/18

delihus:
It was moved here https://github.com/ros2/example_interfaces/tree/jazzy/msg

PR #116, opened 2020-05-18
Add deprecation messages to all messages now in example_interfaces

tfoote:
Followup to #89

### DIFF
--- a/std_msgs/msg/UInt8MultiArray.msg
+++ b/std_msgs/msg/UInt8MultiArray.msg
@@ -1,3 +1,8 @@
+# This was originally provided as an example message.
+# It is deprecated as of Foxy, it is recommended to create your own
+# semantically meaningful message.
+# However if you would like to continue using this please use the equivalent in example_msgs.
+
 # Please look at the MultiArrayLayout message definition for
 # documentation on all multiarrays.
 
--- a/std_msgs/msg/UInt8MultiArray.msg
+++ b/std_msgs/msg/UInt8MultiArray.msg
@@ -1,6 +1,6 @@
 # This was originally provided as an example message.
-# It is deprecated as of Foxy, it is recommended to create your own
-# semantically meaningful message.
+# It is deprecated as of Foxy
+# It is recommended to create your own semantically meaningful message.
 # However if you would like to continue using this please use the equivalent in example_msgs.
 
 # Please look at the MultiArrayLayout message definition for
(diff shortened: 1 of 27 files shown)


std_msgs/msg/UInt8MultiArray.msg
```diff
@@ -1,3 +1,8 @@
+# This was originally provided as an example message.
+# It is deprecated as of Foxy, it is recommended to create your own
```
hidmic, line 2: @tfoote consider
```suggestion
# It is deprecated as of Foxy. 
# It is recommended that you create your own
```
same elsewhere.

tfoote:
It's been in ROS 1 that we generally recommend against using almost any message in std_msgs except Header for a long time for stability.
From the Overview: http://wiki.ros.org/std_msgs

```
However, these types do not convey semantic meaning about their contents: every message simply has a field called "data". Therefore, while the messages in this package can be useful for quick prototyping, they are NOT intended for "long-term" usage. For ease of documentation and collaboration, we recommend that existing messages be used, or new messages created, that provide meaningful field name(s).

Note that this package also contains the "MultiArray" types, which can be useful for storing sensor data. However, the same caveat applies: it's usually "better" (in the sense of making the code easier to understand, etc.) when developers use or create non-generic message types (see discussion in this thread for more detail).
```

However from the naming it leads to a lot of confusion.([For example](https://github.com/ros2/test_interface_files/pull/4#issuecomment-617790928)) And we want to take the opportunity in ROS 2 to avoid that ambiguity. The earlier we do it in this cycle the easier and less disruption it will cause.

std_msgs/msg/UInt8MultiArray.msg
tfoote: Thanks that reads better, updated.

jacobperron:
I guess we shouldn't have run CI for this change since it caused a test failure downstream: https://ci.ros2.org/view/nightly/job/nightly_linux_release/1548/testReport/ros2interface.src.ros2.ros2cli.ros2interface.test/test_cli/test_cli/

See a fix in https://github.com/ros2/ros2cli/pull/516

ros-discourse:

https://discourse.ros.org/t/soss-a-whole-new-approach-to-your-ros-1-ros-2-bridge/17040/18

ros-discourse:

https://discourse.ros.org/t/discussion-standardizing-ros-interfaces/33693/1

PR #123, opened 2020-06-03
Finish up API documentation

brawner:
This adds API documentation to each message and service that didn't already have them.

### DIFF
--- a/trajectory_msgs/msg/JointTrajectoryPoint.msg
+++ b/trajectory_msgs/msg/JointTrajectoryPoint.msg
@@ -2,8 +2,25 @@
 # or positions[, effort] for the trajectory to be executed.
 # All specified values are in the same order as the joint names in JointTrajectory.msg.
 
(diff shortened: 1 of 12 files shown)


trajectory_msgs/msg/JointTrajectoryPoint.msg
```diff
@@ -2,8 +2,25 @@
 # or positions[, effort] for the trajectory to be executed.
 # All specified values are in the same order as the joint names in JointTrajectory.msg.
 
+# Single DOF joint positions for each joint relative to their "0" position.
+# The units depend on the specific joint type: radians for revolute and
+# continuous joints and meters for prismatic joints.
```
Blast545, line 7: The repeated "and" makes it a bit confusing to read, maybe something like:
```suggestion
# Single DOF joint positions for each joint relative to their "0" position.
# The units depend on the specific joint type: radians for revolute and
# continuous joints while meters for prismatic joints.
```
(or a comma separation as the one below this)

PR #125, opened 2020-07-01
Missing cstring header for memcpy in fill_image.hpp

j-rivero:
While running the ABI checker on this package it complains about memcpy not being declared. I'm including here the <cstring> header and use the std namespace.

### DIFF
--- a/sensor_msgs/include/sensor_msgs/fill_image.hpp
+++ b/sensor_msgs/include/sensor_msgs/fill_image.hpp
@@ -36,6 +36,7 @@
 #ifndef SENSOR_MSGS__FILL_IMAGE_HPP_
 #define SENSOR_MSGS__FILL_IMAGE_HPP_
 
+#include <cstring>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
@j-rivero Please merge PRs with squashing in the future,

I just saw that the PR was targeting the `foxy` branch. Can you clarify why and make sure the patch get ported to the default branch right away.

j-rivero:
> @j-rivero Please merge PRs with squashing in the future,

Ok.

> I just saw that the PR was targeting the `foxy` branch. Can you clarify why and make sure the patch get ported to the default branch right away.

It is targeting `foxy` to be able to run the abi-checker on Foxy packages as part of the scope of Quality Level declarations. I'll send another PR to the master branch.

dirk-thomas:
> It is targeting `foxy` to be able to run the abi-checker on Foxy packages as part of the scope of Quality Level declarations.

In the future please always target the latest branch and backport from there.

PR #130, opened 2020-09-02
Update ColorRGBA.msg

NikolasE:
Setting the alpha channel to 1 should help a lot of people who wonder why their marker is not visible in RVIZ

### DIFF
--- a/std_msgs/msg/ColorRGBA.msg
+++ b/std_msgs/msg/ColorRGBA.msg
@@ -1,4 +1,4 @@
 float32 r
 float32 g
 float32 b
-float32 a
+float32 a 1


tfoote (changes requested):
As much as this might help potential new users who forget to set one of 4 fields, with such a core message we need a strong case as to why the change is worth more than the cost. It will require code audits and a transition plan to make sure that we don't suddenly turn a lot of use cases where the default transparent settings are desired. How do we make a transition without breaking all those users? Or is the benefit larger than those costs?

The immediate problem might make sense to leverage the rviz warning box/message to indicate that all markers (or N markers) are alpha zero since they're unrenderable.

I'm going to mark this as request changes. As it needs a much stronger justification to be considered for merging.

clalancette:
No response to the reviewer question in almost 2 years, so I'm going to close this out.  Feel free to reopen and provide answers if you'd like to move forward with this.

PR #135, opened 2020-11-25
Update QDs to QL 1

brawner:

### DIFF
--- a/geometry_msgs/QUALITY_DECLARATION.md
+++ b/geometry_msgs/QUALITY_DECLARATION.md
@@ -2,9 +2,9 @@ This document is a declaration of software quality for the `geometry_msgs` packa
 
 # `geometry_msgs` Quality Declaration
 
-The package `geometry_msgs` claims to be in the **Quality Level 2** category.
+The package `geometry_msgs` claims to be in the **Quality Level 1** category.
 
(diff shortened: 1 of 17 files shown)


geometry_msgs/QUALITY_DECLARATION.md
```diff
@@ -82,15 +82,15 @@ There are no currently copyrighted source files in this package.
```
chapulina, line 82: I noticed that only `4.v` is listed inside testing. Maybe we should list all items?

(the same goes for several other packages)

chapulina (commented):
Bookkeeping - these PRs need to go in first:

* https://github.com/ros2/rcl_interfaces/pull/116
* https://github.com/ros2/rosidl_defaults/pull/15

geometry_msgs/QUALITY_DECLARATION.md
brawner: They would all say the same thing. This is an interfaces package and doesn't require them. It seemed simpler to just write that at the top of section 4

chapulina: Fair enough, sounds good

chapulina (approved):
LGTM, but let's try to merge the PRs in order to make sure we're not missing anything.

PR #137, opened 2020-11-25
Add coverage/performance to qd for sensor_msgs

brawner:
This will also need to be backported to foxy


chapulina (approved):
Nice catch!

PR #139, opened 2020-12-16
Fix PointCloud2Iterator namespacing in docs

andre-nguyen:
The class PointCloud2Iterator isn't in the `msg` namespace, fixed the example in the comments


clalancette (approved):
Looks good to me.  As it is just a change to the documentation, I'm going to take the PR job as sufficient for this and skip CI.  Thanks for the contribution!

PR #141, opened 2021-02-22
Add point cloud message wrapper

niosus:
Before we start: this is a slightly larger PR but it provides the complete functionality. If there are any suggestions on a better process for submitting this, I am absolutely open to any suggestions.

## What does this PR add and why

This PR adds a class `PointCloudMsgWrapper` along with its iterators and typedefs: `PointCloud2View` and `PointCloud2Modifier`. It aims to provide similar functionality to already existing `PointCloud2Modifier` and `PointCloud2Iterator` with more emphasis on the usability and avoiding undefined behavior. At its core, it provides guarantees of correctness for reading/writing to/from point cloud messages if the user is willing to pay two tradeoffs:
- of using a point type that is value constructable and has a comparison operator
- and of potentially having slight gaps in the serialized point cloud due to padding

It is _not_ a replacement to the current solution. It provides an arguably better way to read/write clouds in some conditions discussed below. All the messages written with the proposed classes will be readable by the iterators already available in the repository to date.

The core design principles here were:
- The classes must be easy and intuitive to use
- The access to the underlying data must incur no overhead
- The classes should fail early in case the user makes a mistake, e.g. trying to read a double value where float was stored
- There should be no way to incur undefined behavior using these classes
- The code should allow for relatively simple extension should the default functionality not prove to be enough for power users

Let me illustrate the benefits on an example below.

#### Example comparison to the status quo
Consider an example usage of the existing iterators and modifiers to fill the message from `points` vector:

**Current state:**
```cpp
sensor_msgs::PointCloud2 cloud;
sensor_msgs::PointCloud2Modifier mod{cloud};
mod.resize(N);
mod.setPointCloud2Fields(
  2,
  "x", 1, sensor_msgs::msg::PointField::FLOAT32,
  "y", 1, sensor_msgs::msg::PointField::FLOAT32);
sensor_msgs::PointCloud2Iterator<float> it_x{cloud, "x"};
sensor_msgs::PointCloud2Iterator<float> it_y{cloud, "y"};
for (std::size_t idx = 0U; idx < points.size(); ++idx) {
  *it_x = points[idx].x;
  *it_y = points[idx].y;
  ++it_x;
  ++it_y;
}
```

**Proposed:**
The same can be concisely written with the proposed library:
```cpp
sensor_msgs::PointCloud2 cloud;
PointCloud2Modifier<PointT> modifier{cloud};
for (const auto& point : points) {
  modifier.push_back(point);
}
```

Note, that the points can also be queried from the `PointCloud2Modifier` or `PointCloud2View` through iterators that are able to return a (potentially const) reference to a `PointT` and/or typical vector-like notation, e.g., `view[i]` returns a (potentially const) reference to a `PointT`.  Note that the code has no overhead for accessing the points, is easier to read and provides additional safety guarantees for the user that will be discussed below.

#### User pitfalls in the current approach
Let me outline the core differences between the two approaches on the example provided above.

In the existing iterators and modifiers examples there are the following pitfalls that the user must be mindful of:
- The message must be created with the fields provided in exactly the correct form
- The iterators must then be provided the exact names of the fields configured before. Note also, that these operations, while happening in the same function above, might be happening in different parts of the code and even on different machines, making it the responsibility of the user to match the names and other parameters
- The message must be resized in order to be filled with iterators - there is AFAIK no modifying iterator capable of changing the message size on demand
- The user must make sure the iterator expects to read/write a given type (`float` in the example above). If this type does not match the one actually stored in the `data` field of the message the execution enters the undefined behavior land and might read/write garbage or in the best case crash. This was the initial reason for the code proposed here. Debugging undefined behavior is extremely costly.

#### How the proposed code addresses these pitfalls
With these pitfalls in mind, let me walk you through how the proposed code deals with these and explain the design decisions made in the process:
- The fields are generated automatically given a point struct/class through the modifying constructor of the `PointCloudModifier<PointT>(msg, frame_id)`. This is done by the virtue of `FieldGenerator`s that are able to check if the point struct has a member or a method with a given name and add an appropriate field to the vector of fields stored in the message. As C++ has no reflection, there is no way to write a generic `FieldGenerator`, so we opt for making a tuple of these configurable and provide a sane default value for this tuple that should satisfy most of the users.
- The iterators now return the `PointT` when asked for underlying data, removing the possibility of a mistake on the user side both for reading and writing the data
- The message does not have to be explicitly resized prior to using it with the proposed classes. It has to only be initialized with a point type. An iterator is capable of adding new points to the message and the  `PointCloudModifier` has a familiar `push_back(const PointT&)` method, again, reducing the amount of details that the user must keep track of.
- If we create a `PointCloudView<PointT>(msg)` of a message with fields not matching those generated from the `PointT`, the user will get an exception telling which fields are wrong/missing. This also accounts of type sizes and offsets which might cause an issue on different architectures as theoretically the structs might be generated in different ways. These classes are not designed to be used in this setup and old way can be used to read such messages instead.

### Tradeoffs
I want to explicitly talk about all the pitfalls I can think of in the usage of the proposed solution. They are a few, but I think the benefits are big enough to make it a good tradeoff.

- Every time a wrapper or a view is created a check is performed on the `fields` of the message and the ones generated from the point type to make sure these match. This check takes time but I would argue that this is a one-off operation. A wrapper would be created once per message and read and written many times. Reads and writes are zero-overhead.
- The point is written to the message as a direct memory mapping. Therefore, we might lose some space if the point struct/class has padding. This also is not portable in the general sense of the word. Other systems/compilers might generate a different layout for the point. We do not aim to provide a solution for such a case, but we do provide a check that fails explicitly if the layouts do not match. This also makes messages written by current iterators/modifiers potentially unreadable by the ones proposed here. The other direction will work though.
- There is a limited number of field generators provided. We made a small default tuple of field generators and provide a convenient way to create new ones. The user that is not satisfied with the default must therefore create their own field generators and pass those as a template parameter to the `PointCloudMsgWrapper`. We believe this makes the default usage easy and provides enough freedom to power users.
- The code for generating fields from a point struct/class is slightly convoluted. As any template-heavy code in C++ there is only that much one can do about it. I have made an effort to document all the code and to make it as readable as I could. This code should only be looked at by the experts, and I hope it is readable enough for that purpose.

Thanks for your time and please treat this as a start of a discussion. I am absolutely happy to hear and incorporate any feedback.


niosus:
@dejanpan here is the PR for visibility.

ros-discourse:

https://discourse.ros.org/t/proposal-for-a-new-pointcloud2iterator-pointcloud2modifier/19122/1

sensor_msgs/include/sensor_msgs/point_cloud_msg_wrapper.hpp
```diff
@@ -0,0 +1,496 @@
...
+  class Iterator;
+
+  ///
+  /// @brief      Constructor that wraps a point cloud message. Checks if the message fields
+  ///             correspond to the ones generated from the PointT type and throws in case of a
+  ///             mismatch.
+  ///
+  /// @param      cloud_ref  A reference to the wrapped cloud.
+  ///
+  explicit PointCloudMsgWrapper(CloudMsgT & cloud_ref)
```
niosus, line 158: Just to start a discussion. Should we allow an unsafe version of this wrapper? We could wrap it into a separate class like `IKnowWhatImDoing` and allow the user to manually specify the fields for those cases when this class is otherwise not flexible enough. The plus point - more power to the **power**-user (normal power user would use a custom struct, thus the bold text). The minus point, I can't imagine when this would be needed. We should be able to create a point struct/class with correct alignment/offsets by clever using of `alignas` etc. Thoughts?

sensor_msgs/include/sensor_msgs/point_cloud_msg_wrapper.hpp
```diff
@@ -0,0 +1,496 @@
...
+    }
+    // The idea is the following:
+    // 1. Allocate storage as big as a point.
+    // 2. Fill the storage with zeros.
+    // 3. Iterate through all fields and set the appropriate memory to ones.
+    // 4. Flip all bits - now, if all fields are present, only the padding is non-zero.
+    // 5. Reinterpret this memory as a point.
+    // 6. Compare this point to a value initialized one.
+    // If these points are equal - we have all fields. Otherwise, we are missing a field.
+    alignas(PointT) std::uint8_t buffer[sizeof(PointT)];
```
niosus, line 384: Just to start a discussion: this is one of the most strange parts in this code. Basically, this is invented for the purposes of this class and I am not aware of other usages just like this one. Why? Because we need to make sure that we cover all the fields in case there is a point struct that has field that have no corresponding generators passed into this class. Any ideas on how to make this simpler or how we can get rid of this are welcome. Otherwise, any suggestions on how to test this even better are also much appreciated.

sensor_msgs/include/sensor_msgs/field_generators.hpp
```diff
@@ -0,0 +1,170 @@
...
+/// generated by those.
+///
+/// @tparam     PointT       Type of point
+/// @tparam     FieldAdders  A tuple of field adder types, generated by
+///                          LIDAR_UTILS__DEFINE_FIELD_GENERATOR_FOR_MEMBER
+///
+/// @return     All generated PointField structs.
+///
+template<class PointT, class FieldAdders>
+inline sensor_msgs::msg::PointCloud2::_fields_type generate_fields_from_point()
```
niosus, line 160: A conversation starter: this is the meat of the field generator. This calls a recursive function that generates all the fields that are needed from the point structure. Is this clear enough? Can this be improved? Any thoughts are welcome.

niosus (commented):
Just to jump-start the discussion, here are some comments in the (arguably) most interesting parts of the code. Please don't be scared of the scariest bits and check out the tests if you do. The tests should show off the nice interface :wink:

niosus:
Update: following the discussion [in Autoware](https://gitlab.com/autowarefoundation/autoware.auto/AutowareAuto/-/merge_requests/862/diffs#note_520981557) I removed the custom iterators and use the standard ones from a vector. This seems to work just fine and all the tests are still passing.

xmfcx:
> Update: following the discussion [in Autoware](https://gitlab.com/autowarefoundation/autoware.auto/AutowareAuto/-/merge_requests/862/diffs#note_520981557) I removed the custom iterators and use the standard ones from a vector. This seems to work just fine and all the tests are still passing.

@niosus  I have forked the repository and added this PR as remote and will start working on reviewing it.

niosus:
Thanks @xmfcx 

Just as a heads up, I just realized there is no `resize` method in my class, so I will be adding this. And also I want to add a function to check if the wrapper _can_ be created without failing so that we could pick a wrapper based on the point type. Everything else is a fair game to comment/change.

sensor_msgs/include/sensor_msgs/point_cloud_msg_wrapper.hpp
```diff
@@ -0,0 +1,496 @@
...
+{
+using DefaultFieldGenerators = std::tuple<
+  field_x_generator,
+  field_y_generator,
+  field_z_generator,
+  field_id_generator,
+  field_ring_generator,
+  field_intensity_generator>;
+
+// TODO(igor): this is a general trait, should it live elsewhere?
```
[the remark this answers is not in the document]
niosus: Moved to a different file.

niosus:
@xmfcx just a heads up. I think I'm done with this PR for now. I also had to rebase, so in case you are basing on this branch you would have too, sorry about that. Looking forward to your comments, suggestions, and contributions.

sensor_msgs/include/sensor_msgs/point_cloud_msg_wrapper.hpp
```diff
@@ -0,0 +1,455 @@
...
+/// @tparam     PointT           Type of point to use for message reading/writing.
+/// @tparam     CloudMsgT        Type of point cloud message.
+/// @tparam     FieldGenerators  A tuple of all field generators that allow generating Field structs
+///                              from members of a PointT struct. See
+///                              LIDAR_UTILS__DEFINE_FIELD_GENERATOR_FOR_MEMBER for more details.
+///                              The class provides a sane default value here, but a custom tuple
+///                              can be generated by the user by using
+///                              LIDAR_UTILS__DEFINE_FIELD_GENERATOR_FOR_MEMBER if needed.
+///
+template<typename PointT, typename CloudMsgT, typename FieldGenerators, std::size_t UPPER_BOUND>
```
xmfcx, line 103: What is the reason behind having `CloudMsgT` as a template?

niosus: Mostly to avoid code duplication for `const` and non-`const` versions of the point cloud type.

sensor_msgs/include/sensor_msgs/point_cloud_msg_wrapper.hpp
xmfcx: The current state of this code only checks if the sizes of fields match the size of fields in the given PointType struct right?

Because flipping bits of a 0(uint32_t) is same as flipping bits 2 x 0(uint16_t)s.

Like if PointType had:
uint32_t x;

and fields were like:
- uint16_t x
- uint16_t y

It would pass this test as correct even though they don't match right?

xmfcx: Also following for loop in the code would give a segfault if the fields were accounting for more memory space than the PointType:

Example if 
- fields had x,y,z,intensity for 4x4=16 bytes
- PointType had only x,y,z for 4x3=12 bytes (assume no padding)

```cpp
    alignas(PointT) std::uint8_t buffer[sizeof(PointT)];
    std::fill(buffer, buffer + sizeof(PointT), 0U);
    for (const auto & field : generated_fields) {
      const auto field_first_byte = field.offset;
      const auto field_last_byte =
        field.offset + field.count * static_cast<std::size_t>(sizeof_field(field.datatype));
      std::fill(
        &buffer[field_first_byte],
        &buffer[field_last_byte],
        std::numeric_limits<std::uint8_t>::max());
    }
```

here the `std::fill` will try to access 13th element of buffer which has size 12 and segfault.

niosus: > It would pass this test as correct even though they don't match right?

this particular part - yes, but I do (at least I hope) all the checks that account for these situations before this check starts.

>  the code would give a segfault if the fields were accounting for more memory space than the PointType

I don't think it will. The generated fields are generated _from the same `PointType`_. These are not the fields that are provided by the user in any way. 

You can try to reproduce the failures you suggest by adding a certain point type to the test set. Do the ones you suggest fail the test?

xmfcx: I have added the test to here:

https://github.com/xmfcx/common_interfaces/commit/9cfa2376d4fda69ec0af81ead41086126a294baa

and it fails with following output:
```cpp
PointCloudMsgWrapperTest.point_with_2_custom_fields
30: /home/mfc/projects/ros2_ws/src/common_interfaces/sensor_msgs/test/test_point_cloud_msg_wrapper.cpp:283: Failure
30: Expected: CustomCloudModifier(msg, "some_frame_id") doesn't throw an exception.
30:   Actual: it throws.
30: unknown file: Failure
30: C++ exception with description "Generated fields don't match the members of the point struct. Make sure you have correct generators tuple passed into this class." thrown in the test body.
30: [  FAILED  ]
```

I'm not sure what is the best way to link my changes here since I cannot push into your remote and you cannot push into mine.

niosus: @xmfcx I in the code you linked you are using a different point structure (`PointWithCustomField`) from the one you define. So for now it makes sense that it fails. Please change it to `PointCustomAB` and check again.

I think what we could do is you could open a PR into my branch, this way I can merge your changes directly into this branch. Would that work for you?

xmfcx: Also do you think [BOOST_FUSION_ADAPT_STRUCT](https://www.boost.org/doc/libs/1_68_0/libs/fusion/doc/html/fusion/adapted/adapt_struct.html) could be used here as suggested in [here](https://stackoverflow.com/a/17660354)?

Not sure if we are allowed to use boost here but it would make generating the fields less macro-y for this code.
More like it would offload the macro parts to boost and we could just use `boost::fusion::for_each()` to iterate through the fields of the point struct and generate fields on the runtime. 
And we wouldn't need user entering anything to generate the fields.

xmfcx: > @xmfcx I in the code you linked you are using a different point structure (`PointWithCustomField`) from the one you define. So for now it makes sense that it fails. Please change it to `PointCustomAB` and check again.

Ah my bad, sorry :P
Fixed it and now works.

> I think what we could do is you could open a PR into my branch, this way I can merge your changes directly into this branch. Would that work for you?

Ok, made the PR.

niosus: I am not too familiar with boost fusion. I'm not sure what is the policy of Boost usage here, so I will not target this in my comment. I do not think we can directly use it here though as I just don't see how this could be implemented without an input from the user in the form of generators or alike. For it to work without user input we would need reflection-like capabilities and C++ does not have that unfortunately. But if you know of a way to generate the fields entries from a random struct at runtime - I am all ears.

niosus: Thinking more about this, it might make sense to do smth _like_ boost does. Basically, we would wrap every struct that the user might provide (or instruct the user to do it on their own) and pass the resulting iterable object to our class where we can generate the fields from it etc. The difference here is that the user would need to do it for every type that they might want to use. The up side is that the interface seems to be quite clean. I might have a deeper look into this.

clalancette: We have a policy of no boost in the core, so we shouldn't make any of this dependent on boost.  Alternatively, if you find you really need to use boost, you can make this code a third-party package that depends on boost.

xmfcx: I was thinking if we could use `BOOST_FUSION_ADAPT_STRUCT` in the template typename the wrapper takes or something. But with no-boost policy, it can't be helped.

niosus: @christophebedard yep, that is exactly what I expected, thanks for pitching in!

xmfcx:
https://github.com/niosus/common_interfaces/pull/1/commits/39923f776536f3d8a0c5ace5bbd30beedcce74e3

Here I have replaced usages of `m_cloud_ref.point_step` with `sizeof(PointT)` because latter is defined in compile time thus probably faster. Also it is more local.

xmfcx:
https://github.com/niosus/common_interfaces/pull/1/commits/2b41c86a269c253a814acb83bfa3d9d005ca414b

Here I replaced the resize+assign operation with byte insertion for push_back (named it push_back_2) and in the bench I've added to the test for `100000000` points it performs with about 50ms faster.
```
30: [ RUN      ] PointCloudMsgWrapperTest.push_back_time_test
30: push_back took: 776.124 ms.
30: push_back_2 took: 723.962 ms.
30: [       OK ] PointCloudMsgWrapperTest.push_back_time_test (2213 ms)
```

xmfcx:
> @xmfcx are these commits part of some PR? If yes, could you tag me there so that I could review all changes in one go?

They are part of this PR: https://github.com/niosus/common_interfaces/pull/1

clalancette (commented):
Given that this is all brand-new code, and wraps the existing `PointCloud2` data type, are there any major roadblocks to making this a separate package?  That is, I'm wondering if this could be hosted elsewhere and just have a dependency on `common_interfaces`, rather than putting this into the core.

niosus:
@clalancette nothing stops us from doing that and that might be a very fair point. The main reason for putting this here is that the current way of iterating the `PointCloud2` type is kinda error prone, so it would be nice to have an alternative present already in core. But again, I am very open to suggestions.

clalancette:
> @clalancette nothing stops us from doing that and that might be a very fair point. The main reason for putting this here is that the current way of iterating the `PointCloud2` type is kinda error prone, so it would be nice to have an alternative present already in core. But again, I am very open to suggestions.

There's a balance to be struck here.  On the one hand, ROS aims to ship with "common" things in the core.  So that argues for putting more things here.

On the one hand, the more things we put into the core, the harder it becomes to maintain.  Compile times go up, test times go up, and the overall maintenance burden goes up with every line of code we put into the core.

Most of the time the way we resolve this dilemma is to say that things go into the core after they have become largely agreed upon in the community.  That is, if we find that a large portion of the community is using a certain message, or a certain library, then that is a good candidate to put into the core.

In this particular case, I would say that since this is new code, it seems like it should start its life in a separate repository and package.  If it turns out that it is very widely used, then we can always re-evaluate the decision and think about bringing it into the core.

At least, that is my thinking about it.  @ros2/team other opinions welcome here.

niosus:
@clalancette that does sound very reasonable. Especially considering the cost of failure is quite high, i.e. if we do miss something in implemetation/review it is not that simple to fix this, is this right? We might then want to align with @JWhitleyWork and @xmfcx as to reverting back to the idea of letting it live either in Autoware or we could just release this as a standalone repo under @ApexAI or alike (@dejanpan, what do you say?). So there are options. 

I personally would still prefer to see it live in the core for the reasons I described above but I also completely understand the considerations on the @ros2/team side.

sensor_msgs/include/sensor_msgs/point_cloud_msg_wrapper.hpp
niosus: @dejanpan oh no! Sorry @christophebedard I meant to mention @clalancette here :man_facepalming:  Sorry for the confusion guys.

dejanpan:
> @JWhitleyWork and @xmfcx as to reverting back to the idea of letting it live either in Autoware or we could just release this as a standalone repo under @ApexAI or alike (@dejanpan, what do you say?). So there are options.

It makes sense to release as a standalone and then graduate it into `common_interfaces` if it gets used a lot. We have this package https://gitlab.com/ApexAI/apex_containers for example which is already released and @carlossvg knows how to release it.

@niosus @clalancette

niosus:
The code is now available in @Apex.AI namespace here: https://gitlab.com/ApexAI/point_cloud_msg_wrapper 

I will also soon release it through rosdistro. With this I will close this PR.

FYI @dejanpan @wjwwood @xmfcx @clalancette

Issue #142, opened 2021-03-10
add sensor_msgs_py to Foxy


flynneva:
## Feature request
add sensor_msgs_py to Foxy

#### Feature description
Foxy will be the default distro for awhile now and the `sensor_msgs_py` has already been merged into the master branch for this repo. I was just wondering if it would be possible to also release it for the foxy branch as well.

#### Implementation considerations
should just be a PR/merge I think. ive tested it and it compiles just fine

hidmic:
It is fairly uncommon for a feature (let alone an entire package) to be backported to a previous distro. Though I wouldn't expect this particular addition to compromise downstream code in any way. @jacobperron what do you think about a Foxy backport?

flynneva:
@hidmic, completely agree - however for the most part this `sensor_msgs_py` package is required if a ROS2 user would like to work with Pointcloud2 messages in python i would say.

for that reason i think its pretty critical to get it released for Foxy since most people will be using Foxy probably until the end of the year if not into 2022 I would say.

you can follow the discussion of this package in the PR that added it to ROS2: https://github.com/ros2/common_interfaces/pull/128

there is some debate where it should reside...but that i think is a problem that could be addressed after its available "somewhere".

flynneva:
if you'd like to see how this package is used, you can check out the example I added too that uses it: https://github.com/ros2/examples/pull/276

jacobperron:
Sounds okay to me. I don't expect any disruption related to releasing the new packages into Foxy.

@flynneva Would you like to open a PR to the `foxy` branch?

flynneva:
opened here #143

jacobperron:
Done in #143

PR #143, opened 2021-03-16
[foxy] backport sensor_msgs_py

flynneva:


flynneva:
@jacobperron @hidmic ready for review i think

flynneva:
closes #142

hidmic:
@flynneva it'd be best to cherry-pick the commits to be backported, so as to not lose the connection between both branch histories.

flynneva:
@hidmic i think i did that right?

let me know if I should also cherry-pick the version changes as well - probably want different versions for foxy, no? I wasnt sure.

hidmic:
> @hidmic i think i did that right?

Looks like it. Thanks.

>  probably want different versions for foxy

That's correct. And now I realize that the package version was set to match that of the other packages in `Rolling`. Bloom will not like that version discrepancy, and artificially downgrading this package seems wrong. @clalancette @jacobperron we don't have a policy for this, do we?

clalancette:
> Yeah, that's unfortunate. I don't think we have precedence for something like this. 

Yeah, this is the first time I can remember this issue coming up.

> Moving it to it's own repo would solve the issue, but perhaps that's a bit drastic. Downgrading the version to match the other packages in this repo wouldn't be the end of the world. I'm open to other suggestions.

I wouldn't put this in a separate repository; at least, we had that discussion when this PR first came in, and we decided to keep it here.  While downgrading the version is a bit weird, I don't see why it wouldn't work.  Also, since the foxy and the master branch will never be recombined, I don't see any conflicts arising from it in the future.  Finally, by setting the version number back for Foxy, it maintains the monotonically increasing property of the version numbers (that is, the version number in Foxy will be strictly less than the version number in Rolling/Galactic).  So I'll suggest we just set it back.

flynneva:
so would setting it to match the current foxy `sensor_msgs` version make sense then?

`2.0.3` i think.

hidmic:
> so would setting it to match the current foxy sensor_msgs version make sense then?

> 2.0.3 i think.

That's right @flynneva.

tfoote:
Process note: Downgrading the version if it has not been released is sub optimal for introspection/discoverability/communication. However if it has already been released we can't do it, because the older higher version will take precedence if already installed or cached anywhere in the release pipeline. I think we're in the first case so this is probably ok.

clalancette:
> Process note: Downgrading the version if it has not been released is sub optimal for introspection/discoverability/communication. However if it has already been released we can't do it, because the older higher version will take precedence if already installed or cached anywhere in the release pipeline. I think we're in the first case so this is probably ok.

In this particular case, this is two different release streams.  That is, we've already released this package as `ros-rolling-sensor-msgs-py 2.2.1`.  This PR is backporting it to Foxy, where it will be released as `ros-foxy-sensor-msgs-py 2.0.3` (well, it will probably get bumped for the release, but you get my point).  Since we don't have a foxy -> {rolling,galactic} direct upgrade path, they are on different streams essentially forever.  Thus, I don't think there are any apt problems we need to worry about.

Also, the number in Foxy (2.0.3) will be lower than that in Rolling/Galactic (2.2.1), so there should be no "head-scratching" about the version numbers.

flynneva:
so......can we merge this thing? 😄

hidmic:
Not before CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13975/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8806/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11675/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14058/)

jacobperron (approved):

jacobperron:
Windows warnings are unrelated.

flynneva:
@jacobperron is there anything else needed to get this released to foxy - or will this now be included in some sort of automated release for the entire 'common_interfaces' repo?

jacobperron:
@flynneva A maintainer should make a release for Foxy for it to be included in the next sync.

I will make sure it get's released before the next patch release (there's not tentative date for it, but I will try to do one within a few weeks from now).

flynneva:
@jacobperron  no rush! but good to know. thanks again for everyone's effort on this, really appreciated 😄

PR #153, opened 2021-06-25
Marker textures

gbalke:
This is a pre-cursor to a corresponding update to the RViz triangle marker message system. It allows for a texture to be applied to triangles using UV coordinates.

The functionality has been demonstrated here:
https://github.com/gbalke/rviz_shader_tester


clalancette:
Just so you know, this is quite a large change.  In particular, putting this in will basically invalidate all rosbags that were taken with previous ROS distributions, and it ensures that Humble will be wire-incompatible with Galactic (cross-distribution compatibility isn't something we guarantee, but it is nice to keep when we can).

All of that said, we do allow for this kind of breakage in the core if the change has broad enough applicability.  Can you describe more about UV coordinates, and why we might want them here?

visualization_msgs/msg/Marker.msg
```diff
@@ -56,6 +56,16 @@ geometry_msgs/Point[] points
...
 
+# Only used if the marker has a UV mapping defined for it or one is defined using uv_coordinates.
+# A URI to where the texture is accessible.
+string texture_map
+# Only used when a texture_map is defined and the type specified has some use.
+# Location of each vertex within the texture map; in the range: [0.0-1.0]
+UVCoordinate[] uv_coordinates
+# Only used when a texture_map is defined.
+# Causes the rendering engine to reload the given texture if already loaded.
+bool texture_update
```
sloretz, line 67: Maybe the texture could be reloaded every time a marker with the texture is added? If not, does one marker's `texture_update=True` cause all markers using the same texture to get updated when the texture updates?

tfoote (changes requested):
Overall I think that this is a good extension of the message. However we need to make sure that the message remains self contained.

I know that often textures are maintained in parallel with meshes and are reused. But keeping things self contained is super valuable in this context. Alternatively we should be setting up a parallel stream possibly. But that's getting complicated again.

visualization_msgs/msg/Marker.msg
gbalke: I'm just going to answer this question based on my implementation (which can change depending on preference).

To answer @sloretz,
I currently load textures under their URI into Ogre. This means that if two different markers use the same URI, they will pull the same loaded texture. This also means if one updates, they will both now be using the same updated texture. This can change by appending some metadata to the name used to cache the texture in Ogre such that updating that specific marker without touching others.

To answer @tfoote,
The textures can be of arbitrary size. I have no idea how large people might want to make textures but I think that we can expect up to 4k images (which may be an unreasonable assumption). I imagine for really large surfaces, people will likely just use meshes instead of markers so maybe we can make an assertion about performance loss when textures exceed a certain size and have them be included in the message in their raw format. I was using a URI and caching it in Ogre to minimize data transmission assuming people might want to use very large images for their textures (no idea what the application might be :shrug:).

With regards to modification,
The reason why you would want a separate "modify" param is that the texture may be quite large and you want that to be a one-time cost. Reloading every time a marker with the texture is added could be a way to handle it but I think it's scenario dependent so it's really up to if we want to make that decision for them or let them decide for themselves with the `texture_update` param.

visualization_msgs/msg/Marker.msg
```diff
@@ -56,6 +56,16 @@ geometry_msgs/Point[] points
 # NOTE: alpha is not yet used
 std_msgs/ColorRGBA[] colors
 
+# Only used if the marker has a UV mapping defined for it or one is defined using uv_coordinates.
+# A URI to where the texture is accessible.
+string texture_map
```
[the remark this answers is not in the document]
gbalke: Hmmm I agree but I don't know if it's a good idea to store the image in the message due to how large the image may be (especially if sent with every message). Another strategy could be to have the textures defined separately via another message stream and then the marker system can just apply those textures via a unique identifier present in both messages? This would also allow the removal of `update_texture` as that will not be a constant stream (only sends messages when changes are to be made or a new texture should be loaded) :thinking:.

visualization_msgs/msg/UVCoordinate.msg
```diff
@@ -0,0 +1,4 @@
+# Location of the pixel as a ratio of the width of a 2D texture.
+# Values should be in range: [0.0-1.0].
+float64 u
```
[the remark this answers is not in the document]
gbalke: It's unlikely that we really need a float64. I don't think we can expect images in the billions of pixels hahaha.

visualization_msgs/msg/Marker.msg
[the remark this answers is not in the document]
jacobperron: +1 for having the texture data in a message somewhere that can be recorded.

If we're worried about the marker message being too large because of the texture, then I think a separate texture stream makes sense (either a message or service used to associated a texture with some ID).

jacobperron:
FWIW, it might be worth taking a look at [mesh_tools](https://github.com/uos/mesh_tools) for inspiration. They appear to have message definitions for [textures](https://github.com/uos/mesh_tools/blob/master/mesh_msgs/msg/MeshTexture.msg) and [UV coordinates](https://github.com/uos/mesh_tools/blob/master/mesh_msgs/msg/MeshVertexTexCoords.msg).

gbalke:
> FWIW, it might be worth taking a look at [mesh_tools](https://github.com/uos/mesh_tools) for inspiration. They appear to have message definitions for [textures](https://github.com/uos/mesh_tools/blob/master/mesh_msgs/msg/MeshTexture.msg) and [UV coordinates](https://github.com/uos/mesh_tools/blob/master/mesh_msgs/msg/MeshVertexTexCoords.msg).

I'm happy to go this route if the uri is going to be replaced by an alternate message stream. I think [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier) implies a specific implementation which we don't necessarily want. A unique name will work. I think `texture_id` with a comment describing this should do the trick?

visualization_msgs/msg/Marker.msg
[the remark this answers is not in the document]
gbalke: I started working down the route of texture via an alternate stream. This ended up being a lot more complicated than anticipated for a number of reasons:

1. If we want RViz to apply a texture by topic (a marker specifies a topic which contains the texture to be used), the number of topics will grow quickly which doesn't scale well.
2. If we want to define a static `/texture` topic, we have to figure out a way to guarantee that RViz receives all published messages per `texture_id` before it is started.
3. There are race conditions to applying the textures though this can be mitigated as Ogre has the ability to check for a texture identifier (so can just wait until it has been loaded and ignore otherwise).

My preference is to stick with the URI method. We already have a history of using URIs to load visual messages without the ability to record (meshes). When the resource is unavailable, the mesh handler just throws a warning and does not show anything. In the case of markers and unloaded textures, the marker will still render based on the defined color values just without an applied texture. In addition, I can add the warning regarding the inability to load the given URI. I see this as a fine fallback behavior and it can be stated clearly in the message definition.

At a later date, we can add support for record-able messages via a more complex URI handler that parses a `ros_texture_topic://` prefix or something of the like.

tfoote:
Great find @jacobperron We should definitely leverage the work that mesh_tools has already demonstrated. As they've shown we can encapsulate the texture relatively easily in a message. Then everything will be self contained and there's no issues with synchronization, access, or id mapping.

visualization_msgs/msg/Marker.msg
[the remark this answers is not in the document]
gbalke: I noticed that `mesh_tools` implemented the [texture loading as a service](https://github.com/uos/mesh_tools/blob/e6f9a8cb69644c756a262f626ffe32118437b38c/rviz_map_plugin/src/MeshDisplay.cpp#L901) which is definitely doable just a matter of if the added complexity is the correct approach.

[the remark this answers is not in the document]
tfoote: > My preference is to stick with the URI method. We already have a history of using URIs to load visual messages without the ability to record (meshes). When the resource is unavailable, the mesh handler just throws a warning and does not show anything. In the case of markers and unloaded textures, the marker will still render based on the defined color values just without an applied texture. In addition, I can add the warning regarding the inability to load the given URI. I see this as a fine fallback behavior and it can be stated clearly in the message definition.

Just because we've taken the shortcut in the past doesn't mean it's good to do again. This is something that I remember being a common question on ROS Answer. (Why are my meshes all white?) And I think that the only things that we currently use meshes on are generally urdfs which also have URI style lookups and the workaround that we used with them is to do things like publish them on latched topics.

The service at least is reasonable for providing the same abstraction level. However, I think I need to understand the use cases we're trying to cover here. Are we talking about completely static textures that are reused over and over again. Or are we talking about dynamic textures that are generated based on runtime content. 

Note that we had a previous instance of this for tranforms and a transform server and we optimized to use numeric frame ids instead of strings. And it's much more efficient, but requiring the lookup service to resolve frame names completely compromised the experience. 

Coming into this I think that the basic consideration is that everything should be contained in the message. And it's only if that causes larger overhead that we should then consider increasing the complexity of the system via remote calls or caching. 

I see that there's potential use cases ranging from drawing a TV with live video feed on the surface. To drawing blades of grass with a realistic green pattern. What will be the overhead cost of resending the data in each message versus providing parallel infrastructure to send, store, and query for the data?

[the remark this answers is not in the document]
wjwwood: > This is something that I remember being a common question on ROS Answer. (Why are my meshes all white?)

But this has to do with URDF, which I think is a bit different because the meshes are also specified by URI (not embedded) so the issue would be "I don't see any meshes". If their meshes are white, that means the textures the meshes reference cannot be located, which is a different issue and wouldn't be solved by embedding the mesh data somewhere in a message. I think because of this you're always going to have issues with locating resources for playing back old data, and using `file://` URIs isn't the only problem.

I've encountered the white meshes thing when tf wasn't working, which for some reason would result in a bundle of meshes at the origin, all white. So there might be other reasons for this error case.

> And I think that the only things that we currently use meshes on are generally urdfs which also have URI style lookups and the workaround that we used with them is to do things like publish them on latched topics.

When do we publish meshes used by URDF on topics? Is that common practice, because I've never seen it. Can a URDF even reference a topic?

Also, Marker messages already have URI for meshes. When @gbalke asked me about this I suggested sticking with the URI approach since it would be consistent with that, see:

>  The format is the URI-form used by resource_retriever, including the package:// syntax.

-- https://wiki.ros.org/rviz/DisplayTypes/Marker#Mesh_Resource_.28MESH_RESOURCE.3D10.29_.5B1.1.2B-.5D

Plus I think that if you support URIs then later you could add support for textures and/or meshes over topics or services.

The topics are complicated to implement and ensure textures are received by rviz (durability transient-local isn't a ready solution imo) and the services based solution is not record-able anyway (at this time).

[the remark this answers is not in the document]
gbalke: @tfoote I don't think we can actually decide for the user what the best approach is in this scenario. There's a reason to send the raw image in the message (for minimum latency/synchronization in a video streaming example) and there's reason to have it in a request format (very large textures for high fidelity environments). I think the best option is to provide a method to do all of the above. I think the URI method is still the best method due to the ability to extend to meet needs down the road.

In the case of wanting distributed resources, the user can simply specify a URL or file per the default URI behavior. We may extend this behavior to support two additional modes:
`raw_image_(format)://` followed by an encoded image.
`texture_<topic or id>://` followed by the topic or unique ID to be loaded (implementation similar to mesh_tools).

With this, we can state the performance and logging (rosbag) trade-offs that come along with each method. The benefit of this is the flexibility for the user to decide as I don't think we'll be able to "one-size-fits-all" this scenario.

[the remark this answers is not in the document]
gbalke: I would be happy to add `sensor_msgs/image` if not for the issue of sending the texture every time adding a significant amount of overhead. I also don't think we can avoid this else we have a race condition where the visualizer has to be started (and listening) before the first message is sent else miss the message containing the image information.

visualization_msgs/msg/Marker.msg
gbalke: Removed texture update following the decision to change to texture in message.

gbalke:
After chatting with @tfoote, we came to the decision to include the texture in the message. This was done to provide an ease-of-access to apply textures. While sub-optimal for certain use-cases, this will be recordable and be quick to bring up. For more optimized solutions, more complicated packages may be used such as `mesh_utils`.

visualization_msgs/msg/Marker.msg
```diff
@@ -56,6 +56,11 @@ geometry_msgs/Point[] points
 # NOTE: alpha is not yet used
 std_msgs/ColorRGBA[] colors
 
+# An image to be loaded into the rendering engine as the texture for this marker.
+sensor_msgs/Image texture
```
wjwwood, line 60: Is this the right type to use here? Doesn't this require the texture to be an uncompressed image? Is that a reasonable restriction? We don't need to do image processing on the texture (as far as I know), we're just passing it to the 3D engine to render.

Furthermore, I don't think we don't need a `frame_id` or arguably the timestamp from the `header`, nor the `width` or `height`.

I think `sensor_msgs/CompressedImage` is probably a better fit:

https://docs.ros.org/en/api/sensor_msgs/html/msg/CompressedImage.html

wjwwood (changes requested):
I'm concerned with the approach selected. It deviates from the already existing `MESH_RESOURCE` marker type and it is not future proof, in the case that you want to use another method of specifying the texture.

I would suggest these fields instead:

```
# If texture_uri either empty string ("") or something like "texture_embedded://",
# then the texture can be read from texture_image, otherwise it is loaded via
# resource_retriever, just like mesh_resource.
string texture_resource
sensor_msgs/CompressedImage texture_image
UVCoordinate[] uv_coordinates
```

With code like:

```c++
if (
  marker_msg.texture_resource.empty() ||
  marker_msg.texture_resource == "texture_embedded://")
{
  load_texture_from_compressed_image(marker_msg.texture_image);
} else {
  // load_texture_from_resource_retriever(marker_msg.texture_resource);
  throw std::runtime_error("not implemented");  // or some similar error condition
}
```

That way if/when you want to support loading via `package://` or `file://`, you don't have to change the message definition again.

visualization_msgs/msg/Marker.msg
tfoote: :+1:  Switching to compressed makes sense for this.

gbalke:
The latest changes match @wjwwood's suggestions. I've also added a mesh field at the request of @tfoote, @clalancette, @wjwwood to give meshes the same ability to be recorded in rosbag if one chooses to do so. Implementation will not immediately be available but this is a preemptive change to minimize number of commits modifying this message.

visualization_msgs/msg/Marker.msg
```diff
@@ -56,8 +56,13 @@ geometry_msgs/Point[] points
 # NOTE: alpha is not yet used
 std_msgs/ColorRGBA[] colors
 
+# Texture resource is a special URI that can either reference a texture file in
+# a format acceptable to libcurl or:
```
tfoote, line 60: We should add to the documentation references to resource retriever and use it like we already do for the meshes. (This lets us use package:// syntax too)

https://index.ros.org/p/resource_retriever/

Then for the embedded approach I might suggest `embedded://name` or `attached://name` and then we can mirror it and used the same keyword for the mesh.

visualization_msgs/msg/Mesh.msg
```diff
@@ -0,0 +1,3 @@
+# Used to send raw mesh files.
+string format
```
tfoote, line 2: Expectations of this string should be documented. Is it a list of file types? 

For example see the Compressed image which lists the expected formats: https://github.com/ros2/common_interfaces/blob/d08d14194a108f8626fd05292961703df19374aa/sensor_msgs/msg/CompressedImage.msg#L10

This is important so that we know what standards to use and don't end up with both `dae` and `Collada` trying to express the same thing one in the abberviated suffix vs the full format name.

visualization_msgs/CMakeLists.txt
```diff
@@ -27,6 +28,8 @@ set(msg_files
   "msg/Marker.msg"
   "msg/MarkerArray.msg"
   "msg/MenuEntry.msg"
+  "msg/Mesh.msg"
```
tfoote, line 31: I think that we should clarify that this is not a Mesh natively but an `EncodedMesh` or `MeshFile` that you're going to have to parse it after reading it. I'm not sure what the best modifier is for meshes.

In a parallel way to we're sending a CompressedImage and you're going to need to uncompress it. Versus just reading the image.

tfoote (commented):
This is looking good from our discussion. Structurally I think it's basically there. I mostly want to improve the documentation to make sure that we and everyone following are on the same page.

visualization_msgs/msg/Mesh.msg
wjwwood: Yeah, I'm kind of thinking we should just have one field be a file name, and if it's generated (i.e. no real file) then they can put something like `generated.dae`.

wjwwood (commented):
I second @tfoote's comments and also I think generally we could use more detailed docs, because I find myself asking things like:

- what URI are acceptable
- What is the "texture_name" used for, supposed to be?

And generally the Mesh docs could be updated to match.

visualization_msgs/CMakeLists.txt
gbalke: Changed to `MeshFile`!

visualization_msgs/msg/Mesh.msg
gbalke: Changed to `filename` and added a comment stating it should contain the file extension.

tfoote (approved):

visualization_msgs/msg/Marker.msg
```diff
@@ -56,9 +56,25 @@ geometry_msgs/Point[] points
 # NOTE: alpha is not yet used
 std_msgs/ColorRGBA[] colors
 
+# Texture resource is a special URI that can either reference a texture file in
+# a format acceptable to (resource retriever)[https://index.ros.org/p/resource_retriever/]
+# or an embedded texture via a string matching the format:
+#   "embedded://texture_name"
+string texture_resource
+# An image to be loaded into the rendering engine as the texture for this marker.
+# This will be used iff texture_resource is set to texture_embedded.
```
wjwwood, line 65: This needs to be updated I think.

wjwwood (approved):
One doc fix up needed, otherwise lgtm

visualization_msgs/msg/Marker.msg
gbalke: Fixed! Good catch :smile:

gbalke:
Complete CI via `--packages-above-and-dependencies`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15075/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9807/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12760/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15242/)

gbalke:
CI after merging (IDL changes made last build unstable):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15077/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9809/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12762/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15244/)

PR #167, opened 2021-12-01
Add prism type to the SolidPrimitive.msg (#166)

xmfcx:
This feature adds a `geometry_msgs::msg::Polygon` field to the [shape_msgs/msg/SolidPrimitive.msg](https://github.com/ros2/common_interfaces/blob/master/shape_msgs/msg/SolidPrimitive.msg) in order to represent `PRISM` primitive too.

Explained in detail in #166

Issue #166, opened 2021-12-01
Support Prism type within the SolidPrimitive.msg


xmfcx:
## Feature request

**PR:** https://github.com/ros2/common_interfaces/pull/167/files

#### Feature description

This feature adds a `geometry_msgs::msg::Polygon` field to the [shape_msgs/msg/SolidPrimitive.msg](https://github.com/ros2/common_interfaces/blob/master/shape_msgs/msg/SolidPrimitive.msg) in order to represent `PRISM` primitive too.

<img src="https://user-images.githubusercontent.com/10751153/144294428-9770016b-7c82-4ab8-8b1e-7a98331e6c7a.png" height="300">

#### Implementation considerations

Another way would be to make this its own separate message, since the other shapes (`BOX`, `SPHERE`, `CYLINDER`, `CONE`) don't use the `polygon` field. But since it doesn't use any bandwidth if not used, I thought it'd be ok.

The prism would be constructed from the [`polygon`](http://docs.ros.org/en/api/geometry_msgs/html/msg/Polygon.html) and the `dimensions[PRISM_HEIGHT]`.
- The polygon would be extruded on Z+ and Z- by half of the height.
- Only `x` and `y` fields of the points ([`geometry_msgs/Point32`](http://docs.ros.org/en/api/geometry_msgs/html/msg/Point32.html) type) are used
- Points of the polygon are ordered counter-clockwise.

xmfcx:
I'm closing since this is merged and backports to `foxy` and `galactic` aren't possible due to API/ABI incompatibility. Thanks everyone!

PR #169, opened 2021-12-09
Backport feature Prism in SolidPrimitive.msg to Galactic

xmfcx:
Related Issue: https://github.com/ros2/common_interfaces/issues/166

With https://github.com/ros2/common_interfaces/pull/167 is merged to master, I would like to backport this small feature to galactic.

Since this only adds a new field and type, it should maintain backwards compatibilty.


clalancette:
Similarly, we cannot backport this to Galactic as this would break API/ABI.  So I'll close this out.

clalancette:
Good call, see https://github.com/ros/rosdistro/pull/31428

PR #170, opened 2021-12-09
Backport feature Prism in SolidPrimitive.msg to Foxy

xmfcx:
Related Issue: https://github.com/ros2/common_interfaces/issues/166

With https://github.com/ros2/common_interfaces/pull/167 is merged to master, I would like to backport this small feature to foxy.

Since this only adds a new field and type, it should maintain backwards compatibilty.


clalancette:
> [](https://github.com/ros2/common_interfaces/issues/166)
> 
> With #167 is merged to master, I would like to backport this small feature to foxy.

Unfortunately, we cannot backport this to Foxy.  This would break all existing users of Foxy, as it is not ABI or API compatible.  So I'm going to close this out.

PR #172, opened 2022-01-24
Add YUV420 and YUV444 to image encodings

hemalshahNV:
https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/pixfmt-yuv-planar.html

Adding V4L2_PIX_FMT_NV12 (YUV420) and V4L2_PIX_FMT_NV24 (YUV444) formats to image encodings to facilitate hardware accelerator support.

### DIFF
--- a/sensor_msgs/include/sensor_msgs/image_encodings.hpp
+++ b/sensor_msgs/include/sensor_msgs/image_encodings.hpp
@@ -100,7 +100,7 @@ const char YUV422_YUY2[] = "yuv422_yuy2";
 // YUV 4:2:0 encodings with an 8-bit depth
 // NV21: https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/pixfmt-yuv-planar.html
 const char NV21[] = "nv21";
-// YUV 4:4:4 encodings with 8-bit depth 
+// YUV 4:4:4 encodings with 8-bit depth
 // NV24: https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/pixfmt-yuv-planar.html
 const char NV24[] = "nv24";
 
@@ -194,7 +194,7 @@ static inline int numChannels(const std::string & encoding)
 
   if (encoding == YUV422 ||
     encoding == YUV422_YUY2 ||
-    encoding == NV21 || 
+    encoding == NV21 ||
     encoding == NV24)
   {
     return 2;
@@ -258,7 +258,7 @@ static inline int bitDepth(const std::string & encoding)
 
   if (encoding == YUV422 ||
     encoding == YUV422_YUY2 ||
-    encoding == NV21 || 
+    encoding == NV21 ||
     encoding == NV24)
   {
     return 8;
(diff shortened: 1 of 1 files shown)


clalancette:
@hemalshahNV First thing; can you please sign-off on your commits to make the DCO bot happy?

Besides that, I ended up pushing 71c3024, which should get rid of any linter warnings.

As far as what to call the new entries, I did a bit of archaeology on the code here, particularly trying to find out where the `YUV422` and `YUV422_YUY2` names came from.

The file https://github.com/ros2/common_interfaces/blob/master/sensor_msgs/include/sensor_msgs/image_encodings.hpp was originally copied over from ROS 1 in commit 5ac1598e3830a949cc33b3d82a555fa7d0ad1a59 from 2017.  At that time, there was only the `YUV422` definition.  Subsequently, the `YUV422_YUY2` definition was added in 2019 in commit d4505b57ed20c49853350705ea72eb6f89b2eafb.

The original ROS 1 code here is in https://github.com/ros/common_msgs/blob/noetic-devel/sensor_msgs/include/sensor_msgs/image_encodings.h .  There, the definition was last touched in https://github.com/ros/common_msgs/commit/7dea9124 , but this was just moving it around.  Prior to that, it was initially added to that repository in https://github.com/ros/common_msgs/commit/0e437968 , way back in 2011.  In neither case is it clear where that name initially came from.

I looked a bit into the OpenCV code (as that is one reasonable place where the names might have come from), but OpenCV mostly relies on FFMPEG for this.  The FFMPEG header file that corresponds most closely here is https://git.ffmpeg.org/gitweb/ffmpeg.git/blob/HEAD:/libavutil/pixfmt.h , which has a lot of definitions.  But none of that closely match what we have in here in the image_encodings.hpp file.

I see three ways we can go here:
1. Use the https://www.fourcc.org encodings, as you've done here.  That is at least *a* standard, and we can just say that the older ones don't quite meet that standard.
2. Use the FFMPEG encodings, as in https://git.ffmpeg.org/gitweb/ffmpeg.git/blob/HEAD:/libavutil/pixfmt.h .  Less accessible, but maybe closer to what we already have.
3. Continue on our own path, defining our own names here.

Personally, I'm fine with option 1 (which is what we have in this PR).  @ros2/team other thoughts here?

gbiggs:
Option 1 sounds good to me, with following fourcc for all encodings in the future. Hopefully we can deprecate the older ones at some point.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16013/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10681/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16341/)

audrow (approved):
Looks good to me!

christianrauch:
Can it be that this PR forgot to update `isColor`? According to the kernel documentation, NV21 (https://www.kernel.org/doc/html/v4.10/media/uapi/v4l/pixfmt-nv12.html) and NV24 (https://www.kernel.org/doc/html/v4.10/media/uapi/v4l/pixfmt-nv24.html) are colour formats.

Currently, I cannot use the `cv_bridge` with these formats. So I am wondering, were these format strings added without actually being used in ROS?

gbiggs:
Thanks for the tip. I added #205 to correct the omission.

christianrauch:
I was playing around with these formats and even after I force `is_dst_color_format` in `cv_bridge` to be true, I cannot convert from or to these formats since the conversion code is not in the database. It looks to me as if those formats were mainly used outside of the ROS ecosystem.

Adding this to `cv_bridge` would be a bit of work, since the NV formats would need to use `cvtColorTwoPlane` instead of `cvtColor`. So we need yet another test for single- or two-plane formats.

PR #155, opened 2021-07-17
Numpy version of point_cloud2.read_points_list and point_cloud2.read_points

dheera:
In the spirit of ROS2 which delivers messages with ndarray / array instead of Python lists, this avoids slow Python iteration over points. In this new implementation there are no python loops over actual PointCloud2 points, everything is vectorized.

It works by simply "typecasting" the PointCloud2 data into a struct array using numpy views and recarrays.

PR #175, opened 2022-03-05
Port pointcloud creation to numpy.

Flova:
This pull request ports the `PointCloud` creation to NumPy. It should massively increase the performance for large point clouds as no iteration over the points in python is needed and the bytes are taken directly from NumPy.
Still needs some work regarding other field types. Feel free to comment on my implementation. Suggestions are welcome.
Cheers

### DIFF
--- a/sensor_msgs_py/sensor_msgs_py/point_cloud2.py
+++ b/sensor_msgs_py/sensor_msgs_py/point_cloud2.py
@@ -99,7 +99,7 @@ def read_points(
     # Check if we want to drop points with nan values
     if skip_nans and not cloud.is_dense:
         # Init mask which selects all points
-        not_nan_mask = np.ones(len(points), dtype=np.bool)
+        not_nan_mask = np.ones(len(points), dtype=bool)
         for field_name in points.dtype.names:
             # Only keep points without any non values in the mask
             not_nan_mask = np.logical_and(
@@ -199,7 +199,7 @@ def dtype_from_fields(fields: Iterable[PointField]) -> np.dtype:
             name = f"unnamed_field_{i}"
         else:
             name = field.name
-        assert not name in field_names, "Duplicate field names are not allowed!"
+        assert name not in field_names, "Duplicate field names are not allowed!"
         field_names.append(name)
 
     # Create a tuple for each field containing name and data type
@@ -231,19 +231,23 @@ def create_cloud(
         # Check if this is an unstructured array
         if points.dtype.names is None:
             assert all(fields[0].datatype == field.datatype for field in fields[1:]), \
-                "All fields need to have the same datatype. Pass a structured NumPy array with multiple dtypes otherwise."
+                "All fields need to have the same datatype. Pass a structured NumPy array \
+                    with multiple dtypes otherwise."
             # Convert unstructured to structured array
             points = unstructured_to_structured(
                 points,
                 dtype=dtype_from_fields(fields))
         else:
             assert len(points.dtype.names) == len(fields), \
-                "The number of fields in the structured NumPy array and the PointFields do not match!"
-            assert all(points.dtype.formats[i] == _DATATYPES[fields[i].datatype].str for i in range(len(fields))), \
-                f"PointField and structured NumPy array do not match data types for all fields! \
-                  Check their order. \n \
-                  The field datatypes are: {','.join(list(map(lambda field: _DATATYPES[field.datatype].str, fields)))} \n \
-                  The NumPy datatypes are: {','.join(points.dtype.formats)}"
+                "The number of fields in the structured NumPy \
+                    array and the PointFields do not match!"
+
+            all_fields_have_matching_datatypes = all(map(
+                lambda i: points.dtype.formats[i] == _DATATYPES[fields[i].datatype].str,
+                range(len(fields))))
+            assert all_fields_have_matching_datatypes, \
+                "PointField and structured NumPy array do not match data types for all fields! \
+                    Check their order and types."
             # TODO check if that is the proper way to rename this
             points.dtype.names = [field.name for field in fields]
     else:
@@ -255,7 +259,8 @@ def create_cloud(
 
     # Handle organized clouds
     assert len(points.shape) <= 2, \
-        "Too many dimensions for organized cloud! Points can only be organized in max. two dimensional space"
+        "Too many dimensions for organized cloud! \
+            Points can only be organized in max. two dimensional space"
     height = 1
     width = points.shape[0]
     # Check if input points where an organized cloud (2D array of points)
--- a/sensor_msgs_py/sensor_msgs_py/point_cloud2.py
+++ b/sensor_msgs_py/sensor_msgs_py/point_cloud2.py
@@ -38,15 +38,14 @@ sensor_msgs/src/sensor_msgs/point_cloud2.py
 """
 
 import array
-import sys
 from collections import namedtuple
-from typing import Iterable, List, NamedTuple, Optional
-
 import numpy as np
 from numpy.lib.recfunctions import (structured_to_unstructured,
                                     unstructured_to_structured)
+import sys
 from sensor_msgs.msg import PointCloud2, PointField
 from std_msgs.msg import Header
+from typing import Iterable, List, NamedTuple, Optional
 
 _DATATYPES = {}
 _DATATYPES[PointField.INT8] = np.dtype(np.int8)
@@ -58,7 +57,7 @@ _DATATYPES[PointField.UINT32] = np.dtype(np.uint32)
 _DATATYPES[PointField.FLOAT32] = np.dtype(np.float32)
 _DATATYPES[PointField.FLOAT64] = np.dtype(np.float64)
 
-DUMMY_FIELD_PREFIX = "unnamed_field"
+DUMMY_FIELD_PREFIX = 'unnamed_field'
 
 
 def read_points(
@@ -92,12 +91,12 @@ def read_points(
     # Keep only the requested fields
     if field_names is not None:
         assert all(field_name in points.dtype.names for field_name in field_names), \
-            "Requests field is not in the fields of the PointCloud!"
+            'Requests field is not in the fields of the PointCloud!'
         # Mask fields
         points = points[list(field_names)]
 
     # Swap array if byte order does not match
-    if bool(sys.byteorder != "little") != bool(cloud.is_bigendian):
+    if bool(sys.byteorder != 'little') != bool(cloud.is_bigendian):
         points = points.byteswap(inplace=True)
 
     # Check if we want to drop points with nan values
@@ -151,7 +150,7 @@ def read_points_numpy(
     :return: Numpy array containing all points.
     """
     assert all(cloud.fields[0].datatype == field.datatype for field in cloud.fields[1:]), \
-        "All fields need to have the same datatype. Use `read_points()` otherwise."
+        'All fields need to have the same datatype. Use `read_points()` otherwise.'
     structured_numpy_array = read_points(
         cloud, field_names, skip_nans, uvs, reshape_organized_cloud)
     return structured_to_unstructured(structured_numpy_array)
@@ -205,8 +204,8 @@ def dtype_from_fields(fields: Iterable[PointField]) -> np.dtype:
         # Datatype as numpy datatype
         datatype = _DATATYPES[field.datatype]
         # Name field
-        if field.name == "":
-            name = f"{DUMMY_FIELD_PREFIX}_{i}"
+        if field.name == '':
+            name = f'{DUMMY_FIELD_PREFIX}_{i}'
         else:
             name = field.name
         # Handle fields with count > 1 by creating subfields with a suffix consiting
@@ -215,10 +214,10 @@ def dtype_from_fields(fields: Iterable[PointField]) -> np.dtype:
         for a in range(field.count):
             # Add suffix if we have multiple subfields
             if field.count > 1:
-                subfield_name = f"{name}_{a}"
+                subfield_name = f'{name}_{a}'
             else:
                 subfield_name = name
-            assert subfield_name not in field_names, "Duplicate field names are not allowed!"
+            assert subfield_name not in field_names, 'Duplicate field names are not allowed!'
             field_names.append(subfield_name)
             # Create new offset that includes subfields
             field_offsets.append(field.offset + a * datatype.itemsize)
@@ -253,16 +252,16 @@ def create_cloud(
         # Check if this is an unstructured array
         if points.dtype.names is None:
             assert all(fields[0].datatype == field.datatype for field in fields[1:]), \
-                "All fields need to have the same datatype. Pass a structured NumPy array \
-                    with multiple dtypes otherwise."
+                'All fields need to have the same datatype. Pass a structured NumPy array \
+                    with multiple dtypes otherwise.'
             # Convert unstructured to structured array
             points = unstructured_to_structured(
                 points,
                 dtype=dtype_from_fields(fields))
         else:
             assert points.dtype == dtype_from_fields(fields), \
-                "PointFields and structured NumPy array dtype do not match for all fields! \
-                    Check their field order, names and types."
+                'PointFields and structured NumPy array dtype do not match for all fields! \
+                    Check their field order, names and types.'
     else:
         # Cast python objects to structured NumPy array (slow)
         points = np.array(
@@ -272,8 +271,8 @@ def create_cloud(
 
     # Handle organized clouds
     assert len(points.shape) <= 2, \
-        "Too many dimensions for organized cloud! \
-            Points can only be organized in max. two dimensional space"
+        'Too many dimensions for organized cloud! \
+            Points can only be organized in max. two dimensional space'
     height = 1
     width = points.shape[0]
     # Check if input points are an organized cloud (2D array of points)
@@ -282,8 +281,8 @@ def create_cloud(
 
     # Convert numpy points to array.array
     memory_view = memoryview(points)
-    casted = memory_view.cast("B")
-    array_array = array.array("B")
+    casted = memory_view.cast('B')
+    array_array = array.array('B')
     array_array.frombytes(casted)
 
     # Put everything together
--- a/sensor_msgs_py/sensor_msgs_py/point_cloud2.py
+++ b/sensor_msgs_py/sensor_msgs_py/point_cloud2.py
@@ -129,8 +129,7 @@ def read_points_numpy(
         uvs: Optional[Iterable] = None,
         reshape_organized_cloud: bool = False) -> np.ndarray:
     """
-    Read equally typed fields from  sensor_msgs.PointCloud2 message
-    as a unstructured numpy array.
+    Read equally typed fields from  sensor_msgs.PointCloud2 message as a unstructured numpy array.
 
     This method is better suited if one wants to perform build math operations
     on e.g. all x,y,z fields.
@@ -187,7 +186,7 @@ def read_points_list(
 
 def dtype_from_fields(fields: Iterable[PointField]) -> np.dtype:
     """
-    Converts a Iterable of sensor_msgs.msg.PointField messages to a np.dtype.
+    Convert a Iterable of sensor_msgs.msg.PointField messages to a np.dtype.
 
     :param fields: The point cloud fields.
                    (Type: iterable of sensor_msgs.msg.PointField)
@@ -217,6 +216,7 @@ def create_cloud(
         points: Iterable) -> PointCloud2:
     """
     Create a sensor_msgs.msg.PointCloud2 message.
+
     :param header: The point cloud header. (Type: std_msgs.msg.Header)
     :param fields: The point cloud fields.
                    (Type: iterable of sensor_msgs.msg.PointField)
--- a/sensor_msgs_py/sensor_msgs_py/point_cloud2.py
+++ b/sensor_msgs_py/sensor_msgs_py/point_cloud2.py
@@ -42,7 +42,7 @@ from collections import namedtuple
 from typing import Iterable, List, Optional, NamedTuple
 
 import numpy as np
-import numpy.typing as npt
+from numpy.lib.recfunctions import structured_to_unstructured
 from sensor_msgs.msg import PointCloud2, PointField
 from std_msgs.msg import Header
 
@@ -62,7 +62,7 @@ def read_points(
         field_names: Optional[List[str]] = None,
         skip_nans: bool = False,
         uvs: Optional[Iterable] = None,
-        reshape_organized_cloud: bool = False) -> npt.ArrayLike:
+        reshape_organized_cloud: bool = False) -> np.ndarray:
     """
     Read points from a sensor_msgs.PointCloud2 message.
 
@@ -73,7 +73,7 @@ def read_points(
                       (Type: Bool, Default: False)
     :param uvs: If specified, then only return the points at the given
         coordinates. (Type: Iterable, Default: None)
-    :return: Numpy array containing all points.
+    :return: Structured NumPy array containing all points.
     """
     assert isinstance(cloud, PointCloud2), \
         'Cloud is not a sensor_msgs.msg.PointCloud2'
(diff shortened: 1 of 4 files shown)


aprotyas:
Repeating my comments from https://github.com/ros2/geometry2/pull/507:

> CI is failing here because it couldn't resolve the `python-numpy` rosdep key for Ubuntu Jammy. I believe the dependency needs to be on `python3-numpy` instead.

> Also, it looks like your commits aren't signed-off. Do you mind signing them off?
https://docs.ros.org/en/rolling/Contributing/Developer-Guide.html#change-control-process

sensor_msgs_py/sensor_msgs_py/point_cloud2.py
```diff
@@ -144,40 +145,6 @@ def read_points_list(cloud, field_names=None, skip_nans=False, uvs=[]):
                                                 skip_nans, uvs)]
 
 
-def create_cloud(header, fields, points):
```
aprotyas, line 147: Isn't this part of the public API? This shouldn't be removed, unless I'm missing something.

Flova: You are right. I just added the code I used locally. I will adapt it to satisfy all of the existing interfaces.

Flova:
Why is the DCO failing? Github shows my commits as correctly signed.

sensor_msgs_py/sensor_msgs_py/point_cloud2.py
Flova: I added it again.

aprotyas:
> Why is the DCO failing? Github shows my commits as correctly signed.

It looks like you signed the commits with your GPG key, but you need to add a Signed-off-by line in your commit message. 

Check here: https://docs.ros.org/en/rolling/Contributing/Developer-Guide.html#id5

sensor_msgs_py/sensor_msgs_py/point_cloud2.py
```diff
@@ -67,56 +71,60 @@ def read_points(cloud, field_names=None, skip_nans=False, uvs=[]):
...
+
+    # Cast bytes to numpy array
+    points = np.ndarray(
+        shape=(cloud.width, len(cloud.fields)),
+        # Limitation: only one datatype per point cloud is possible, not per field
+        dtype=_DATATYPES[cloud.fields[0].datatype][0],
+        buffer=cloud.data)
+
+    # Swap array if byte order does not match
+    if bool(sys.byteorder != "little") ^ bool(cloud.is_bigendian):
```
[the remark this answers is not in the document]
Flova: I don't think Python has a logical
XOR. I could compose one using the existing logical operators, but this works as well and is much shorter and cleaner imo.

sensor_msgs_py/sensor_msgs_py/point_cloud2.py
```diff
@@ -67,56 +71,60 @@ def read_points(cloud, field_names=None, skip_nans=False, uvs=[]):
...
+        field_ids = list(map(
+            lambda field: field[0],
+            filter(
+                lambda field: field[1].name in field_names,
+                enumerate(cloud.fields))))
+
+    # Cast bytes to numpy array
+    points = np.ndarray(
+        shape=(cloud.width, len(cloud.fields)),
+        # Limitation: only one datatype per point cloud is possible, not per field
```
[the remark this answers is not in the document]
Flova: Numpy arrays are only able to contain one datatype. A change that would resolve this issue but breaks the current API would be to return an array for each "collum"/field. This could be implemented by parsing the bytes first as an uint8 or similar so each byte gets one slot in the array. The array would be reshaped so each point gets its own row of bytes. The rows are then casted by the byte size of the fields into a new array with the desired field data type. This is a bit slower than the currently proposed approach, but it should still be significantly faster than the existing implantation.
In addition to that a new API function could be added to query a user defined number of uniform fields as a single array.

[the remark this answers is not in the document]
Flova: Regarding breaking API changes, did this package already work in ros2 due to the PyKDL issue. If not I would be less concerned regarding breaking the API, because people coming from ros1 already face a significant number of breaking changes.

sensor_msgs_py/sensor_msgs_py/point_cloud2.py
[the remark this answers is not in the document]
timonegk: You can simply use `!=`

sensor_msgs_py/sensor_msgs_py/point_cloud2.py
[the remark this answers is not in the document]
timonegk: Maybe line 80 could be changed to an `if` to use the fast implementation if possible and resort to the old implementation otherwise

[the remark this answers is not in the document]
Flova: Oh sorry I confused this PR with the one in `tf2_sensor_msgs`. It indeed does not depend on PyKDL.

I also missed that combinded dtype arrays are indeed possible in numpy as long as the field length stays the same in all over all elements. So I give it a try.

I think NumPy arrays are Iterable so returning a numpy array should comply with the documented API.

Keeping both implemetations would also be an option but I would prefere to have two functions insead of an `if` in this case.

sensor_msgs_py/sensor_msgs_py/point_cloud2.py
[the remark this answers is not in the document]
Flova: Oh right.

sensor_msgs_py/sensor_msgs_py/point_cloud2.py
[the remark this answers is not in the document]
Flova: I adapted the code to support differently typed fields.

Flova:
Duplicate of #155, but this PR has a bit more features and is activly worked on.

Flova:
While the CI still fails with some error related to python3.9 vs python3.10 in Jammy all tests are passing locally on my Ubuntu 20.04 ROS rolling machine.

Flova:
Some performance testing:

As data, I used an unstructured NumPy array containing 480000 points.

Serialization using the old code:
700ms
Serialization using the updated code:
10ms
Deserialization using the old code:
140ms
Deserialization using the new code (read_points_numpy aka with cast to unstructured array):
143 µs
Deserialization using the new code (read_points aka as structured array):
51.7 µs

clalancette:
> While the CI still fails with some error related to python3.9 vs python3.10 in Jammy all tests are passing locally on my Ubuntu 20.04 ROS rolling machine.

Yeah, that's fine.  This is currently broken across the board (we are waiting for some fixes in Jammy itself).  Once we are closer to having this approved, we'll run CI on https://ci.ros2.org to see the status.

Flova:
I plan to write some more tests regarding organized point clouds as well as ones with mixed dtypes.

sensor_msgs_py/sensor_msgs_py/point_cloud2.py
```diff
@@ -157,28 +228,52 @@ def create_cloud(header, fields, points):
...
+    return PointCloud2(
+        header=header,
+        height=height,
+        width=width,
+        is_dense=False,
+        is_bigendian=sys.byteorder != 'little',
+        fields=fields,
+        point_step=points.dtype.itemsize,
+        row_step=(points.dtype.itemsize * width),
+        data=points.tobytes())
```
apockill, line 273: @Flova I just spent a while improving performance for PointCloud2 message creation on the ros2_numpy repo, and thought I'd pass on some knowledge. I noticed that  doing `data=np_array.tobytes()` can be quite slow on large point clouds. I have some code in [this PR](https://github.com/Box-Robotics/ros2_numpy/pull/3) using a memoryview and `array.array.frombytes()` which I found to be 100x faster.

Let me know what you think- there's some overlap in what we're working on I think.

apockill, line 273: @Flova I just spent a while improving performance for PointCloud2 message creation on the ros2_numpy repo, and thought I'd pass on some knowledge. I noticed that  doing `data=np_array.tobytes()` can be quite slow on large point clouds. I have some code in [this PR](https://github.com/Box-Robotics/ros2_numpy/pull/3) using a memoryview and `array.array.frombytes()` which I found to be 115x faster.

Let me know what you think- there's some overlap in what we're working on I think.

Here's an example of how to do the same in this repo:
```suggestion
    memory_view = memoryview(points)
    if memory_view.nbytes > 0:
        array_bytes = memory_view.cast("B")
    else:
        # Casting raises a TypeError if the array has no elements
        array_bytes = b""
    as_array = array.array("B")
    as_array.frombytes(array_bytes)

    # Put everything together
    return PointCloud2(
        header=header,
        height=height,
        width=width,
        is_dense=False,
        is_bigendian=sys.byteorder != 'little',
        fields=fields,
        point_step=points.dtype.itemsize,
        row_step=(points.dtype.itemsize * width),
        data=as_array)
```

Flova: I really liked ros_numpy in ROS1. Cool that it is still a thing in ROS2.

I did some performance testing as seen above. I did not noticed speeds in the seconds as described in your PR. Even for very large point clouds (~7GB), but this could also be related to the used hardware. This being said I also wondered about a multiple magnitude difference in serialisation vs. deserialization. With the serialisation being significantly slower (ms instead of us).

We also kind of do the same thing. You are right with this. But I don't know if the maintainers want to add ros_numpy as a dependency.

clalancette: > We also kind of do the same thing. You are right with this. But I don't know if the maintainers want to add ros_numpy as a dependency.

I'd rather not, both to keep the number of dependencies down, and also to avoid circular dependency problems (it looks like ros2_numpy depends on nav_msgs, which is in this repository).  Can we just use portions of the code that are relevant?

Flova: Okay. I think both ones are very similar now, but I had the idea to make a PR to ros2_numpy to port so minor stuff from here to there (Mainly dropping the `pftype_sizes` mapping as it can be inferred from the dtype).

Flova: I tested the proposed changes on my machine (ROS2 rolling, Ubuntu 20.04, Ryzen 1700X) and could not reproduce the described behavior. It was even slightly slower:

This PR:
```python
points = np.random.rand(1080*1920, 3).astype(np.float32)

In [84]: points = unstructured_to_structured(points, names=["x", "y", "z"])

In [85]: %time a = create_cloud_xyz32(Header(), points)
CPU times: user 9.11 ms, sys: 7.67 ms, total: 16.8 ms
Wall time: 17.2 ms

In [86]: %time a = create_cloud_xyz32(Header(), points)
CPU times: user 1.76 ms, sys: 8.04 ms, total: 9.8 ms
Wall time: 9.78 ms

In [87]: %time a = create_cloud_xyz32(Header(), points)
CPU times: user 4.15 ms, sys: 14 µs, total: 4.16 ms
Wall time: 4.18 ms

In [88]: %time a = create_cloud_xyz32(Header(), points)
CPU times: user 4.17 ms, sys: 18 µs, total: 4.19 ms
Wall time: 4.21 ms

In [89]: %time a = create_cloud_xyz32(Header(), points)
CPU times: user 4.27 ms, sys: 32 µs, total: 4.31 ms
Wall time: 4.33 ms

In [90]: %time a = create_cloud_xyz32(Header(), points)
CPU times: user 4.18 ms, sys: 19 µs, total: 4.2 ms
Wall time: 4.23 ms
```

Proposed changes:
```python
In [5]: points = np.random.rand(1080*1920, 3).astype(np.float32)

In [6]: points = unstructured_to_structured(points, names=["x", "y", "z"])

In [7]: %time a = create_cloud_xyz32(Header(), points)
CPU times: user 12.8 ms, sys: 8.48 ms, total: 21.2 ms
Wall time: 20.5 ms

In [8]: %time a = create_cloud_xyz32(Header(), points)
CPU times: user 11.1 ms, sys: 7.56 ms, total: 18.7 ms
Wall time: 18.6 ms

In [9]: %time a = create_cloud_xyz32(Header(), points)
CPU times: user 156 µs, sys: 12.4 ms, total: 12.6 ms
Wall time: 12.5 ms

In [10]: %time a = create_cloud_xyz32(Header(), points)
CPU times: user 3.86 ms, sys: 12.2 ms, total: 16 ms
Wall time: 15.9 ms

In [11]: %time a = create_cloud_xyz32(Header(), points)
CPU times: user 6.83 ms, sys: 3.31 ms, total: 10.1 ms
Wall time: 10.1 ms
```

Flova: This also happens with different arrays and dtype combinations.

Flova:
```python
    def __init__(self, **kwargs):
        assert all('_' + key in self.__slots__ for key in kwargs.keys()), \
            'Invalid arguments passed to constructor: %s' % \
            ', '.join(sorted(k for k in kwargs.keys() if '_' + k not in self.__slots__))
        from std_msgs.msg import Header
        self.header = kwargs.get('header', Header())
        self.height = kwargs.get('height', int())
        self.width = kwargs.get('width', int())
        self.fields = kwargs.get('fields', [])
        self.is_bigendian = kwargs.get('is_bigendian', bool())
        self.point_step = kwargs.get('point_step', int())
        self.row_step = kwargs.get('row_step', int())
        self.data = array.array('B', kwargs.get('data', []))
        self.is_dense = kwargs.get('is_dense', bool())
```

This is the PointCloud constructor on my installation. Like you described, but without the performance implications.

apockill: The problem is that the `self.data = ...` line is actually calling a property for the data setter:

```python
class PointCloud2:
	...

	@data.setter
    def data(self, value):
        if isinstance(value, array.array):
            assert value.typecode == 'B', \
                "The 'data' array.array() must have the type code of 'B'"
            self._data = value
            return
        self._data = array.array('B', value)
```

^ This is what it looks like for foxy. The `array.array('B, value)` line is quite expensive, at least from my testing. Doing `array.array.frombytes(a_memoryview)` was faster. I'm wondering if `rolling` improved this setter logic to be more efficient for bytes datatypes.

Flova: I agree with you, but my code snippet is sufficient, if you only use the constructor. It also includes the  `array.array('B', value)` part, so it is skipped later on by the setter in the best case. But the interesting part `array.array('B', value)` is included.

apockill: Ahh my apologies, I think I got confused. You're right- your code snippet is running `array.array('B, value)`. The difference is that the implementation for `array.array.frombytes(memoryview)` does a simple `memcpy` for the entire buffer. ([implementation of frombytes](https://github.com/python/cpython/blob/434ffb7f1f86e6b0cdfad3ede59993934d86e464/Modules/arraymodule.c#L1682))

This is different than the `array.array('B', value)` [implementation](https://github.com/python/cpython/blob/434ffb7f1f86e6b0cdfad3ede59993934d86e464/Modules/arraymodule.c#L2630) which is more complicated, and seems to be slower (from my testing). 

Either way, the memoryview implementation is only one copy: the `memcpy`, whereas doing `np.tobytes()` is one copy, and then `array.array('B", value)` performs another copy, and iterates over the data which is added cost.

Flova: I agree that there are happening two copies, but for now I could not measure any performance benefit. Could you try to reproduce your tests describing the exact configuration of your data and system?

apockill: Okay, I had a chance to do some quick tests: https://drive.google.com/drive/u/0/folders/1kxmAn1ND_HF_eCCpH2kXz1V78ajQnP6D

I created a minimal reproducible benchmark that doesn't include ros2_numpy or the PointCloud2 messages. I was curiously unable to reproduce the extremely slow serialization that I saw with ros2_numpy- instead, I saw only a 51% increase.

My theory is that something about how ros2_numpy creates the point cloud ([pointcloud2_to_array](https://github.com/Box-Robotics/ros2_numpy/blob/238b6a2d66d04fb687ce421cc6a7ce63428bbaa3/ros2_numpy/point_cloud2.py#L134)) might be creating an array that's somehow slow to call `tobytes()` on... I think I need to do some more investigating on my end.

 These are the results on my machine:
```
Implementation 1: 2.3597640991210938 seconds
Implementation 2: 1.2081842422485352 seconds
```

Flova: I tested your minimal example, and it was indeed faster. But now comes the funny part. It was exactly equal if I added the `PointCloud2` message again!
I dug deeper and the issue lies in the implementation of the constructor. Which has a redundant typecast. If you use the property directly, it is similar to your implementation.

This therefore seems to be an upstream issue. We should open an issue there and use your proposed changes.

Timing:
```bash
Implementation 1: 2.7484822273254395 seconds   # This PR
Implementation 2: 1.379800796508789 seconds    # Property with your cast
Implementation 3: 2.714930295944214 seconds    # Constructor (as in this PR) with your cast
```

<details>
  <summary>Code</summary>

  ```python
import array
from time import time
from contextlib import contextmanager
from sensor_msgs.msg import PointCloud2, PointField
import numpy as np
import sys
from std_msgs.msg import Header


@contextmanager
def timer(name: str):
    start_time = time()
    yield
    print(f"{name}: {time() - start_time} seconds")


def perform_tests():
    samples = 100
    np_cloud = np.load("np_cloud.npy")
    # dtype: [('x', '<f4'), ('y', '<f4'), ('z', '<f4'), ('a', 'u1'), ('rgb', '<f4')]

    with timer("Implementation 1"):
        for _ in range(samples):
            fields = [PointField(name='x', offset=0,
                         datatype=PointField.FLOAT32, count=1),
              PointField(name='y', offset=4,
                         datatype=PointField.FLOAT32, count=1),
              PointField(name='z', offset=8,
                         datatype=PointField.FLOAT32, count=1)]
            c = PointCloud2(
                header=Header(),
                height=1,
                width=len(np_cloud),
                is_dense=False,
                is_bigendian=sys.byteorder != 'little',
                fields=fields,
                point_step=np_cloud.dtype.itemsize,
                row_step=(np_cloud.dtype.itemsize * len(np_cloud)),
                data=np_cloud.tobytes())

    with timer("Implementation 2"):
        for _ in range(samples):
            fields = [PointField(name='x', offset=0,
                         datatype=PointField.FLOAT32, count=1),
              PointField(name='y', offset=4,
                         datatype=PointField.FLOAT32, count=1),
              PointField(name='z', offset=8,
                         datatype=PointField.FLOAT32, count=1)]
            
            memory_view = memoryview(np_cloud)
            casted = memory_view.cast("B")
            array_array = array.array("B")
            array_array.frombytes(casted)
            
            c = PointCloud2(
                header=Header(),
                height=1,
                width=len(np_cloud),
                is_dense=False,
                is_bigendian=sys.byteorder != 'little',
                fields=fields,
                point_step=np_cloud.dtype.itemsize,
                row_step=(np_cloud.dtype.itemsize * len(np_cloud)))
            c.data = array_array


    with timer("Implementation 3"):
        for _ in range(samples):
            fields = [PointField(name='x', offset=0,
                         datatype=PointField.FLOAT32, count=1),
              PointField(name='y', offset=4,
                         datatype=PointField.FLOAT32, count=1),
              PointField(name='z', offset=8,
                         datatype=PointField.FLOAT32, count=1)]
            
            memory_view = memoryview(np_cloud)
            casted = memory_view.cast("B")
            array_array = array.array("B")
            array_array.frombytes(casted)
            
            c = PointCloud2(
                header=Header(),
                height=1,
                width=len(np_cloud),
                is_dense=False,
                is_bigendian=sys.byteorder != 'little',
                fields=fields,
                point_step=np_cloud.dtype.itemsize,
                row_step=(np_cloud.dtype.itemsize * len(np_cloud)),
                data = array_array)


perform_tests()

```
  
</details>

Comment (Flova) [reply sensor_msgs_py/sensor_msgs_py/point_cloud2.py:292]: I opened #177 and #176 for these issues.

Comment (apockill) [reply sensor_msgs_py/sensor_msgs_py/point_cloud2.py:292]: This is amazing, thank you for the in-depth debugging. I had no idea that `__debug__` was on by default on most machines. That's definitely the cause of the extremely slow serialization I was seeing in https://github.com/Box-Robotics/ros2_numpy/pull/3!

**Review by gbiggs — APPROVED** [review]
LGTM pending passing CI.

**Comment by gbiggs** [comment]
CI:

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/16435/)
- Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/11039/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/16802/)

**Comment by gbiggs** [comment]
CI has picked up a small pile of flake8 linting errors.

```
08:12:16 _________________________________ test_flake8 __________________________________
08:12:16 test/test_flake8.py:38: in test_flake8
08:12:16     assert rc == 0, \
08:12:16 E   AssertionError: Found 20 code style errors / warnings:
08:12:16 E     ./sensor_msgs_py/point_cloud2.py:42:1: I100 Import statements are in the wrong order. 'from collections import namedtuple' should be before 'import sys'
08:12:16 E     ./sensor_msgs_py/point_cloud2.py:61:22: Q000 Double quotes found but single quotes preferred
08:12:16 E     ./sensor_msgs_py/point_cloud2.py:95:13: Q000 Double quotes found but single quotes preferred
08:12:16 E     ./sensor_msgs_py/point_cloud2.py:100:30: Q000 Double quotes found but single quotes preferred
08:12:16 E     ./sensor_msgs_py/point_cloud2.py:154:9: Q000 Double quotes found but single quotes preferred
08:12:16 E     ./sensor_msgs_py/point_cloud2.py:208:26: Q000 Double quotes found but single quotes preferred
08:12:16 E     ./sensor_msgs_py/point_cloud2.py:209:20: Q000 Double quotes found but single quotes preferred
08:12:16 E     ./sensor_msgs_py/point_cloud2.py:218:33: Q000 Double quotes found but single quotes preferred
08:12:16 E     ./sensor_msgs_py/point_cloud2.py:221:54: Q000 Double quotes found but single quotes preferred
08:12:16 E     ./sensor_msgs_py/point_cloud2.py:256:17: Q000 Double quotes found but single quotes preferred
08:12:16 E     ./sensor_msgs_py/point_cloud2.py:264:17: Q000 Double quotes found but single quotes preferred
08:12:16 E     ./sensor_msgs_py/point_cloud2.py:275:9: Q000 Double quotes found but single quotes preferred
08:12:16 E     ./sensor_msgs_py/point_cloud2.py:285:31: Q000 Double quotes found but single quotes preferred
08:12:16 E     ./sensor_msgs_py/point_cloud2.py:286:31: Q000 Double quotes found but single quotes preferred
08:12:16 E     ./test/test_point_cloud2.py:107:10: Q000 Double quotes found but single quotes preferred
08:12:16 E     ./test/test_point_cloud2.py:108:10: Q000 Double quotes found but single quotes preferred
08:12:16 E     ./test/test_point_cloud2.py:109:10: Q000 Double quotes found but single quotes preferred
08:12:16 E     ./test/test_point_cloud2.py:228:51: Q000 Double quotes found but single quotes preferred
08:12:16 E     ./test/test_point_cloud2.py:228:58: Q000 Double quotes found but single quotes preferred
08:12:16 E     ./test/test_point_cloud2.py:228:65: Q000 Double quotes found but single quotes preferred
08:12:16 E   assert 1 == 0
```

Flova:
@gbiggs I hopefully fixed all flake8 errors

gbiggs:
Thanks! Another CI run:

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/16457/)
- Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/11047/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/16809/)

Flova:
I'll definitely need to investigate why my flake8 config is so different to the one used by the CI. But I am now more or less confident that the imports are ordered correctly.

gbiggs:
CI again:

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/16462/)
- Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/11052/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/16814/)

Flova:
Finally ^^

gbiggs:
Thanks for the contribution!

PR #178, opened 2022-03-15
Add sensor_msgs_library target and install headers to include/${PROJECT_NAME}

sloretz:
This adds a modern CMake target `sensor_msgs_library` usable downstream as sensor_msgs::sensor_msgs_library and installs headers to a unique include directory as part of ros2/ros2#1150

It might depend on ros2/rosidl#662, I'm not sure.

### DIFF
--- a/sensor_msgs/CMakeLists.txt
+++ b/sensor_msgs/CMakeLists.txt
@@ -55,8 +55,20 @@ rosidl_generate_interfaces(${PROJECT_NAME}
   ADD_LINTER_TESTS
 )
 
-install(DIRECTORY include/${PROJECT_NAME}/
+rosidl_get_typesupport_target(cpp_target "${PROJECT_NAME}" "rosidl_generator_cpp")
+
+add_library(${PROJECT_NAME}_library INTERFACE)
+target_include_directories(${PROJECT_NAME}_library INTERFACE
+  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
+  "$<INSTALL_INTERFACE:include/${PROJECT_NAME}>")
+target_link_libraries(${PROJECT_NAME}_library INTERFACE
+  "${cpp_target}")
+
+install(DIRECTORY include/
   DESTINATION include/${PROJECT_NAME}
+)
+install(
+  TARGETS ${PROJECT_NAME}_library EXPORT export_${PROJECT_NAME}
 )
 
 if(BUILD_TESTING)
@@ -73,7 +85,12 @@ if(BUILD_TESTING)
   target_link_libraries(test_sensor_msgs "${cpp_typesupport_target}")
 endif()
 
+# Export old-style CMake variables
+ament_export_include_directories("include/${PROJECT_NAME}")
+
+# Export modern CMake targets
+ament_export_targets(export_${PROJECT_NAME})
+
 ament_export_dependencies(rosidl_default_runtime)
-ament_export_include_directories(include)
 
 ament_package()


sloretz:
CI (build: `--packages-above-and-dependencies sensor_msgs` test: `--packages-select sensor_msgs`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16407/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11017/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16780/)

pablogs9:
This makes micro-ROS type support fails because we are not using `rosidl_generator_cpp` at all, and we rely on `rosidl_generator_c`.

Is there any possibility of making this modification optional? Or maybe selecting `rosidl_generator_c` or `rosidl_generator_cpp` depending on which one is found?

I found that just replacing 

```cmake
rosidl_get_typesupport_target(cpp_target "${PROJECT_NAME}" "rosidl_generator_cpp")
```

with

```cmake 
rosidl_get_typesupport_target(cpp_target "${PROJECT_NAME}" "rosidl_generator_c")
```
it works as expected.

Should we assume that cpp generator and/or cpp typesupport is mandatory?

Any idea?

pablogs9:
Another question @sloretz @clalancette , why this has been done only in this `sensor_msgs` package and not in others?

sloretz:
> Another question @sloretz @clalancette , why this has been done only in this sensor_msgs package and not in others?

`sensor_msgs` is the only package that [also provides non-generated C++ headers](https://github.com/ros2/common_interfaces/tree/master/sensor_msgs/include/sensor_msgs) to do things with the generated c++ code.

> Is there any possibility of making this modification optional?

Yes, I think so. The headers in the `sensor_msgs::sensor_msgs_library` target need C++ code to be generated for them, but the target and header installation could be conditioned on that generator being present. I recommend opening a PR that checks `if(TARGET ${cpp_target})` and only installs headers and creates the `sensor_msgs::sensor_msgs_library` target if that check passes. 

> I found that just replacing [...] with [...]

That probably worked because downstream packages using those are depending on `${sensor_msgs_TARGETS}` which also includes the generated c++ target. I wouldn't expect that to work if someone depended on `sensor_msgs::sensor_msgs_library` only.

pablogs9:
> sensor_msgs is the only package that [also provides non-generated C++ headers](https://github.com/ros2/common_interfaces/tree/master/sensor_msgs/include/sensor_msgs) to do things with the generated c++ code. 

As far as I understood `.msg` and other type definition formats provided by ROS 2 are intended to declare a certain type independently of the programming language. Just like IDL. In that sense, you are assuming that this `.msg` and `.srv` are going to have a C++ type support. IMHO this should be divided into a `sensor_msgs` package with type definition and other `sensor_msgs_utils_cpp` with the non-generated C++ code.

> Yes, I think so. The headers in the sensor_msgs::sensor_msgs_library target need C++ code to be generated for them, but the target and header installation could be conditioned on that generator being present. I recommend opening a PR that checks if(TARGET ${cpp_target}) and only installs headers and creates the sensor_msgs::sensor_msgs_library target if that check passes.

I do not understand how I can do `if(TARGET ${cpp_target})`  if `cpp_target` is set in `rosidl_get_typesupport_target(cpp_target "${PROJECT_NAME}" "rosidl_generator_cpp")` and `rosidl_get_typesupport_target` result in a fatal error if `rosidl_generator_cpp` is not available.

In any case, I have proposed an approach here: https://github.com/ros2/common_interfaces/pull/183

RFRIEDM-Trimble:
Is there a reason to not also add an alias library? It's supported in the version of cmake used here.
https://cmake.org/cmake/help/v3.5/command/add_library.html#alias-libraries

PR #189, opened 2022-05-18
Fix SolidPrimitive.msg to contain a single Polygon

xmfcx:
Follow up from https://github.com/ros2/common_interfaces/pull/167

In my previous PR I was supposed to add a single `geometry_msgs::msg::Polygon` to the message but I've added an array of it by mistake :man_facepalming: 

It's even a part of humble now :(

This PR fixes it.

### DIFF
--- a/shape_msgs/msg/SolidPrimitive.msg
+++ b/shape_msgs/msg/SolidPrimitive.msg
@@ -47,4 +47,4 @@ uint8 CONE_RADIUS=1
 # Points of the polygon are ordered counter-clockwise.
 
 uint8 PRISM_HEIGHT=0
-geometry_msgs/Polygon[] polygon
+geometry_msgs/Polygon polygon


xmfcx:
@clalancette the ABI won't be compatible but is it possible to include this fix in `humble` since it came out relatively new?

The worst case scenario I will add a comment line on the old message as a back-port clarification. Something like `// Only add a single polygon here.`

clalancette:
> @clalancette the ABI won't be compatible but is it possible to include this fix in `humble` since it came out relatively new?

Today is the last possible moment to do this.  I'm very much scared of doing this, but we'll discuss it internally and see what we want to do here.

xmfcx:
> Today is the last possible moment to do this. I'm very much scared of doing this, but we'll discuss it internally and see what we want to do here.

I see, I am so sorry :bowing_man: for making this mistake and realizing it too late. Thank you so much for helping.

audrow (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16732/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11315/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17127/)

clalancette:
Besides the CI, I also went through all of the packages released into Humble to see which ones could *potentially* be broken by this.  It's actually a fairly short list of packages that use the SolidPrimitives message, the list looks to be:

* geometric_shapes
* moveit_core
* moveit_ros
* moveit_commander
* ompl
* moveit_msgs
* moveit_visual_tools
* rc_reason_clients
* rc_reason_msgs
* rosbridge_suite

I went through and looked at all of the uses here, and none of them seem to use the polygon or PRISM type.  I also went ahead and built all of them locally with this patch applied, so I think we are OK to go forward with this.  Still waiting on CI to finish, then we'll backport and release.

audrow:
@mergifyio backport humble

PR #192, opened 2022-05-20
retain width and height after resize

tonylitianyu:
The width and height was modified after this resize call. Now the width and height will be retained if they were set before. A new method is also provided to input the set width and height.

Signed-off-by: tonylitianyu <tonylitianyu@gmail.com>

### DIFF
--- a/sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.hpp
+++ b/sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.hpp
@@ -125,14 +125,13 @@ inline void PointCloud2Modifier::reserve(size_t size)
 
 inline void PointCloud2Modifier::resize(size_t size)
 {
+  size_t total_size = size * cloud_msg_.point_step;
+  cloud_msg_.data.resize(total_size);
+
   size_t original_size = cloud_msg_.height * cloud_msg_.width;
   if (original_size == size) {
     return;
   }
-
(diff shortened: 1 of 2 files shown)


clalancette:
Can you please retarget this to the `master` branch?  If and when we decide to take it there, we can then backport it to Foxy.

WilliamLewww:
I think this change is for VIPER which uses Foxy. Would it be okay to do it the other way around? (forward port to master). If not, that would be perfectly fine!

sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.hpp
```diff
@@ -125,12 +125,19 @@ inline void PointCloud2Modifier::reserve(size_t size)
 
 inline void PointCloud2Modifier::resize(size_t size)
 {
-  cloud_msg_.data.resize(size * cloud_msg_.point_step);
+  size_t original_size = cloud_msg_.height * cloud_msg_.width;
+  if (original_size == size) {
```
mallanmba, line 129: I think the return should happen after the `cloud_msg_.data.resize(total_size);` call but before the `// Update height/width`

clalancette:
> I think this change is for VIPER which uses Foxy. Would it be okay to do it the other way around? (forward port to master). If not, that would be perfectly fine!

Our general workflow is to first get things onto master, and then backport into other releases as necessary/requested.  So I'd prefer to keep with that workflow here.

PR #198, opened 2022-07-29
Remove reference to old implementation

Flova:
Remove reference to old implementation and port of said implementation, as the file has been rewritten from scratch since then.

Not high priority, but the reference might cause some confusion and is definitely outdated. Other suggesting regarding a rewording instead of the removal are welcome.


gbiggs (approved):

gbiggs:
LGTM. Thanks for the clean up!

PR #202, opened 2022-08-03
use regex for matching cv types

ijnek:
@clalancette Would you mind reviewing because this is a ``// TODO(clalancette)``?

I'm guessing we should keep ``ABSTRACT_ENCODING_PREFIXES`` there in case it's being used by others downstream.

### DIFF
--- a/sensor_msgs/include/sensor_msgs/image_encodings.hpp
+++ b/sensor_msgs/include/sensor_msgs/image_encodings.hpp
@@ -178,7 +178,7 @@ static inline int numChannels(const std::string & encoding)
   std::cmatch m;
   // ex. 8UC -> 1, 8UC5 -> 5
   if (std::regex_match(encoding.c_str(), m, cv_type_regex)) {
-    return (m[3] == "") ? 1 : atoi(m[3].str().c_str());
+    return (m[3] == "") ? 1 : std::atoi(m[3].str().c_str());
   }
 
   if (encoding == YUV422 ||
--- a/sensor_msgs/include/sensor_msgs/image_encodings.hpp
+++ b/sensor_msgs/include/sensor_msgs/image_encodings.hpp
@@ -33,6 +33,7 @@
 #define SENSOR_MSGS__IMAGE_ENCODINGS_HPP_
 
 #include <cstdlib>
+#include <regex>
 #include <stdexcept>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


sensor_msgs/include/sensor_msgs/image_encodings.hpp
```diff
@@ -172,24 +175,10 @@ static inline int numChannels(const std::string & encoding)
...
-      encoding.substr(
-        prefix.size(), encoding.size() - prefix.size()
-      ).c_str());  // ex. 8UC5 -> 5
-    if (n_channel != 0) {
-      return n_channel;  // valid encoding string
-    }
+  std::cmatch m;
+  // ex. 8UC -> 1, 8UC5 -> 5
+  if (std::regex_match(encoding.c_str(), m, cv_type_regex)) {
+    return (m[3] == "") ? 1 : atoi(m[3].str().c_str());
```
gbiggs, line 181:
```suggestion
    return (m[3] == "") ? 1 : std::atoi(m[3].str().c_str());
```

sensor_msgs/include/sensor_msgs/image_encodings.hpp
```diff
@@ -236,24 +225,10 @@ static inline int bitDepth(const std::string & encoding)
...
-      encoding.substr(
-        prefix.size(), encoding.size() - prefix.size()
-      ).c_str());  // ex. 8UC10 -> 10
-    if (n_channel != 0) {
-      return atoi(prefix.c_str());  // valid encoding string
-    }
+  std::cmatch m;
+  // ex. 8UC -> 8, 8UC10 -> 10
+  if (std::regex_match(encoding.c_str(), m, cv_type_regex)) {
+    return atoi(m[0].str().c_str());
```
gbiggs, line 231:
```suggestion
    return std::atoi(m[0].str().c_str());
```

gbiggs (commented):
Thanks for doing this.

Is there a test that verifies the behaviour is unchanged?

ijnek:
Yes, there were already some tests that verify the behaviour and pass with these changes: https://github.com/ros2/common_interfaces/blob/rolling/sensor_msgs/test/test_image_encodings.cpp

gbiggs (approved):

gbiggs:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17096/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11646/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17514/)

ijnek:
From the build logs, the problem seems that the packages above sensor_msgs aren't being detected in the [Build](https://ci.ros2.org/job/ci_linux/17096/console#console-section-13) step, any idea why..?

gbiggs:
I think I misunderstood how `--packages-up-to` and `--packages-above` interact, and `colcon` took the latter first, meaning `sensor_msgs` didn't have its dependencies available. It will take a little longer, but I'll just have it build the whole workspace so the tests can run.

PR #78, opened 2019-11-15
Add the YUY2 image encoding

sgvandijk:
This adds the YUY2/YUYV version of the YUV 4:2:2 image encoding, as discussed in #76 .

I chose the name `YUV422_YUY2`, to use the most popular fourcc code `YUY2` and try to be consistent with the existing `YUV422` encoding.

Issue #204, opened 2022-09-04
use FourCC codes for image encodings


christianrauch:
## Feature request

The `sensor_msgs/Image` currently uses the `encoding` field to document how the raw data in the `data` field has to be interpreted. The fourcc (4 character code) is a 32bit integer formed by 4 ASCII characters, that uniquely describes the memory layout. 

#### Feature description

The `encoding` field can contain an arbitrary text string, with some predefined strings in `sensor_msgs::image_encodings`. Those strings are not standardised and do not document the memory layout properly. The YUV formats even reference non-official documentation containing advertisements.

Instead of using these non-standard strings, I propose to rely on the fourcc widely used across the kernel and the userspace to uniquely and properly document the memory layout.

#### Implementation considerations

To support fourcc, either a `uint32` or a `char[4]` field `fourcc` should be added to the `sensor_msgs/Image` message definition alongside the old `encoding`. To maintain backward compatibility, the logic could be that if `encoding` is empty, the `fourcc` would take the place as a definition of the memory layout.

Alternatively, the `encoding` string could also contain the fourcc alongside the "legacy" encodings. However, this would require an expensive string matching to test if `encoding` contains a fourcc or a "legacy" format encoding and it must be ensured that there are no collisions.

clalancette:
We had a short discussion about this today internally.

The most important point that came out of that discussion is that we aren't totally sure whether fourcc is the way to go for the encoding.  That is, we aren't sure what pieces of software use fourcc, and what pieces of software use something else.  In particular, we could use references to the kernel code, V4L2 code, OpenCV, and ROS code showing that this standard is actually widely used.  That would allow us to make an easier decision about this.  Thanks.

christianrauch:
- kernel: https://github.com/torvalds/linux/blob/master/include/uapi/drm/drm_fourcc.h
- V4L2: https://github.com/torvalds/linux/blob/v5.19/include/uapi/linux/videodev2.h#L518-L784
- OpenCV: https://github.com/opencv/opencv/blob/4.x/modules/videoio/include/opencv2/videoio.hpp#L1129-L1136
- ROS: As far as I know, FourCC is not used in the ROS ecosystem. The point of this proposal is to introduce FourCC to the ROS ecosystem, starting with the message. 

FourCC is used all over the image/video/media stack to specify the memory layout. We should adapt this and use this specification instead of inventing custom formats that are not properly defined.

Issue #76, opened 2019-11-09
Add YUYV image encoding


sgvandijk:
----
#### Feature description
Most popular USB cameras that are used in mobile/hobby robots, such as the Logitech C9xx cameras, provide images in YUV 4:2:2 format in YUYV order.

There is a `YUV422` encoding defined in `image_encodings.hpp` but that is defined to be in UYVY order.

With an encoding constant defined for YUYV order we can then go on to handle that format in e.g. `cv_bridge` to convert it to other formats so there are no custom solutions needed like [done in the ROS 1 `usb_cam` driver](https://github.com/ros-drivers/usb_cam/blob/develop/src/usb_cam.cpp#L327).

#### Implementation considerations
The implementation will be almost the same as for `YUV422`. The main consideration I think is the naming. Looking at https://www.fourcc.org/yuv.php the canonical name for this format seems to be `YUY2`, with `YUYV` being a synonym. OpenCV indicates the same in its [conversion codes (e.g. `COLOR_YUV2RGB_YUYV = COLOR_YUV2RGB_YUY2`)](https://docs.opencv.org/master/d8/d01/group__imgproc__color__conversions.html).

With that I'd propose either:
1. `YUY2`
2. `YUV422_YUY2` <-- seems best to me
3. `YUV422_YUYV`

I'd be happy to make a PR for this if this is accepted and a name is decided, and I then plan to add handling of it in `cv_bridge` as I'd like to use that in our [V4L2 camera driver](https://gitlab.com/boldhearts/ros2_v4l2_camera).

clalancette:
I'd go for either 1 or 2.  I have a slight preference for 1 (as it is less verbose), but I defer to your judgement on which is clearer.  Please go ahead and open a PR for it; when you do, please be sure to include a link to exactly which pixel format you are talking about.  Thanks!

mjcarroll:
I'm good with either 1 or 2 with a preference towards 2.  Looking forward to the PR!

clalancette:
Fixed by #78, so closing.

PR #214, opened 2023-01-04
update YUV format codes and documentation

christianrauch:
The YUV formats seem to be ambiguous. With this PR, I would like to clarify this and properly document their memory layout. There is no "yuv422_yuy2" format and the "yuv422" format refers to the format code `YU16`. The documentation mentions a "UYUV" format, which does not exist. Additionally, the website http://www.fourcc.org is not reachable anymore.

With the wayback machine, I compared the documented memory layout from http://www.fourcc.org with the one from the Linux kernel documentation from https://www.kernel.org/doc/html/latest/. I found that for the `yuv422` the memory layout described at [fourcc.org](https://web.archive.org/web/20210316221649/https://www.fourcc.org/pixel-format/yuv-uyvy/) matches `V4L2_PIX_FMT_UYVY` at [kernel.org](https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/pixfmt-packed-yuv.html#id1) and [`yuv422_yuy2`](https://web.archive.org/web/20210224214058/https://www.fourcc.org/pixel-format/yuv-yuy2/) matches `V4L2_PIX_FMT_YUYV`.

The NV21 and NV24 formats have unambiguous format codes across the kernel documentation. For the YUV 4:2:2 formats, different sources seem to use different names, hence I propose to use the fourcc instead of an ambiguous name.

See https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/videodev2.h for a list of codes used with video devices.

@sgvandijk Since you added the "yuv422_yuy2" format via https://github.com/ros2/common_interfaces/pull/78, can you comment if my changes make sense to you?

### DIFF
--- a/sensor_msgs/include/sensor_msgs/image_encodings.hpp
+++ b/sensor_msgs/include/sensor_msgs/image_encodings.hpp
@@ -92,17 +92,22 @@ const char BAYER_BGGR16[] = "bayer_bggr16";
 const char BAYER_GBRG16[] = "bayer_gbrg16";
 const char BAYER_GRBG16[] = "bayer_grbg16";
 
-// Miscellaneous
+// YUV formats
 // YUV 4:2:2 encodings with an 8-bit depth
-// UYUV version: http://www.fourcc.org/pixel-format/yuv-uyvy
-const char YUV422[] = "yuv422";
-// YUYV version: http://www.fourcc.org/pixel-format/yuv-yuy2/
-const char YUV422_YUY2[] = "yuv422_yuy2";
+// https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/pixfmt-packed-yuv.html#id1
+// fourcc: UYVY
+const char UYVY[] = "uyvy";
+const char YUV422[] = "yuv422";  // deprecated
+// fourcc: YUYV
+const char YUYV[] = "yuyv";
+const char YUV422_YUY2[] = "yuv422_yuy2";  // deprecated
+
 // YUV 4:2:0 encodings with an 8-bit depth
-// NV21: https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/pixfmt-yuv-planar.html
+// NV21: https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/pixfmt-yuv-planar.html#nv12-nv21-nv12m-and-nv21m
 const char NV21[] = "nv21";
+
 // YUV 4:4:4 encodings with 8-bit depth
-// NV24: https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/pixfmt-yuv-planar.html
+// NV24: https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/pixfmt-yuv-planar.html#nv24-and-nv42
 const char NV24[] = "nv24";
 
 // Prefixes for abstract image encodings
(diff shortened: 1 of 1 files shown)


sgvandijk:
The naming of the format was discussed in https://github.com/ros2/common_interfaces/issues/76. As mentioned there, one of the reasons of adding the format was to be able to then extend cv_bridge to be able to perform the right conversion. OpenCV defines:
```c++
  cv::COLOR_YUV2RGB_YUY2 = 115,
  cv::COLOR_YUV2BGR_YUY2 = 116, 
```
and then the aliases:
```c++
  cv::COLOR_YUV2RGB_YUYV = COLOR_YUV2RGB_YUY2,
  cv::COLOR_YUV2BGR_YUYV = COLOR_YUV2BGR_YUY2,
  cv::COLOR_YUV2RGB_YUNV = COLOR_YUV2RGB_YUY2,
  cv::COLOR_YUV2BGR_YUNV = COLOR_YUV2BGR_YUY2, 
```
so hinting they treat YUY2 as the canonical name, but also YUYV and YUNV as alternative names, seemingly following the table that used to be at fourcc.org.

It is true that this doesn't follow the V4L2 codes and that V4L2 drivers return the code `"YUYV"`,  which I also had to handle that [in my V4L2 camera driver](https://gitlab.com/boldhearts/ros2_v4l2_camera/-/blob/8db2609c52d0b6153ae5c075e0dfe23ca7ae1a33/src/v4l2_camera_device.cpp#L220-221). However I opted to be consistent with OpenCV as the more general framework rather than the more specific V4L2.

FWIW, [Windows seems to use YUY2 as well](https://learn.microsoft.com/en-us/windows/win32/medfound/recommended-8-bit-yuv-formats-for-video-rendering#yuy2) and does not use YUYV as an alias AFAICT. Also see [RFC2361](https://www.rfc-editor.org/rfc/rfc2361).

FFmpeg mixes things a bit: running `ffmpeg -f v4l2 -i /dev/video0` gives me:

>  Stream #0:0: Video: rawvideo (YUY2 / 0x32595559), yuyv422, 640x480, 147456 kb/s, 30 fps, 30 tbr, 1000k tbn, 1000k tbc

so they use the YUY2 FourCC, which they then map to the internal yuyv422 type.

With all that said, I'd agree that YUYV would be clearer and I wouldn't be against adopting that per se, I had that name in mind initially as well. But for almost everywhere I look except for V4L2, YUY2 is the common default name to refer to this encoding.

christianrauch:
Thanks for the detailed response. That means that `YUYV` is used on Linux and `YUY2` is used on Windows, but they both describe the same memory layout. The [ffmpeg pixel format mapping](https://github.com/FFmpeg/FFmpeg/blob/n5.1.2/libavcodec/raw.c#L52-L57) indeed shows that `MKTAG('Y', 'U', 'Y', '2')` and `MKTAG('Y', 'U', 'Y', 'V')` are mapped to `AV_PIX_FMT_YUYV422`. In this case, I am also fine with using `YUY2` as the fourcc.

I wouldn't so much rely on OpenCV for choosing those names. It is more important that they properly specify the precise memory layout so that the image can be reconstructed manually from the raw data. When you get a data stream from the camera with a given fourcc, then the mapping to the ROS "encodings" should be unambiguous (see also https://github.com/ros2/common_interfaces/issues/204).

But in any case, I find the `yuv422` encoding name ambiguous as there are multiple YUV 4:2:2 encodings.

tfoote (approved):
This looks cleaner and more clearly documented.

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18366/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12902/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19036/)

PR #217, opened 2023-03-10
Enable document generation using rosdoc2 for ament_python pkgs

Yadunund:
Fix missing `exec_depends` in `package.xmls` to generate documentation without any warnings using `autodoc`

### DIFF
--- a/sensor_msgs_py/package.xml
+++ b/sensor_msgs_py/package.xml
@@ -14,8 +14,9 @@
   <author email="michel@ekumenlabs.com">Michel Hidalgo</author>
   <author email="sebastian.grans@gmail.com">Sebastian Grans</author>
 
-  <exec_depend>sensor_msgs</exec_depend>
   <exec_depend>python3-numpy</exec_depend>
+  <exec_depend>sensor_msgs</exec_depend>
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
If this is only necessary for doc generation we should use the 'doc_depend' tag.

clalancette:
> If this is only necessary for doc generation we should use the 'doc_depend' tag.

It turns out that it is a dependency here, we just forgot to add it in the past and got lucky since it is a transitive dependency as well.  See https://github.com/ros2/common_interfaces/blob/4bac182a0a582b5e6b784d9fa9f0dabc1aca4d35/sensor_msgs_py/sensor_msgs_py/point_cloud2.py#L45

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18217/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12748/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18886/)

tfoote (approved):
In that case lgtm

PR #218, opened 2023-03-28
Adds ARROW_STRIP to Marker.msg

rr-tom-noble:
ROS2 version of this PR: https://github.com/ros/common_msgs/pull/190

I won't have a chance to work on the Rviz PR to implement the `ARROW_STRIP` functionality for ROS2 until next month (April), so I suppose we can either leave this PR open until then, or merge with a "currently unimplemented" comment.

I've changed the description above the `points` field to include the new type, but have left the `colours` field unchanged as (at least in the ROS1 version of the Rviz PR) this functionality was unimplemented for `ARROW_STRIP`.


tfoote (approved):
+1 for merging into rolling

I think it's safe for backporting too. Publishing of newer data to older implementations will break parsing (hopefully with an unknown datatype type of error) . That should gracefully degrade. If anyone knows of another case we should look into that more if there's strong interest in backporting. I suspect that backporting it will depend on the complexity of the relevant rviz pull request complexity.

clalancette (changes requested):
I think that in order to move forward with this, we'd also want an implementation of the RViz change for RViz2.  Then we can remove the `unimplemented` comment here and merge the two of them together.

rr-tom-noble:
@clalancette

Makes sense. I should have time this week to raise a PR with rviz2. Will update this PR and when merged

rr-tom-noble:
Rviz2 PR partially implemented:
https://github.com/ros2/rviz/pull/972

Currently waiting for answers to some questions

ahcorde:
@clalancette are you happy with this PR ?

clalancette (changes requested):
This should be rebased onto the latest, and all changes to files except for `Marker.msg` dropped.  Then this will be a lot closer to be ready to merge (though I think we should wait for https://github.com/ros2/rviz/pull/972 to be ready before we do that).

ahcorde (changes requested):
Do you mind to rabse this with `rolling` ?

There are many changes not related with this PR

tfoote:
I've rebased this onto rolling in #242 Closing this as superseded.

PR #222, opened 2023-05-11
Fix read_points_numpy field_names parameter

broughtong:
The current `point_cloud2.read_points_numpy` function takes a pointcloud2 message and returns an unstructured numpy array of the points.
There is an assertion test that all fields are the same data type, required for unstructured numpy arrays.
However if you have multiple data types and use the `field_names` parameter to select only a certain subset of fields with a matching data type, all fields are still tested for the same data type and this will fail.
Change fixes this so that if you are using "field_names", only checks the selected fields have matching data types.
Includes a test case to verify that a call with non-uniform field types fails, and a test case for selecting fields with matching types works even when dissimilar types exist in the pointcloud.


tfoote (approved):
Thanks for this clarification. The assertion was overly broad for the necessary protection. And the test cases are good to validate the differentiation.

clalancette:
Did CI get run for this change?

PR #223, opened 2023-05-12
Allow pointcloud create_cloud function to set specific point_step

broughtong:
If I try to convert a pointcloud message to an array, and then use `create_cloud` to convert it back to a message, it will not work if there is padding at the end.

The original lidar data is laid out like this (note the itemsize/padding):

`{'names':['x','y','z','intensity','t','reflectivity','ring','ambient','range'], 'formats':['<f4','<f4','<f4','<f4','<u4','<u2','<u2','<u2','<u4'], 'offsets':[0,4,8,16,20,24,26,28,32], 'itemsize':48}`

The current `create_cloud` will try to reconstruct the point layout from the field info with `dtype_from_fields`, which in this case will take into account field padding, but not the padding at the end. It generates the following (again note the itemsize):

`{'names':['x','y','z','intensity','t','reflectivity','ring','ambient','range'], 'formats':['<f4','<f4','<f4','<f4','<u4','<u2','<u2','<u2','<u4'], 'offsets':[0,4,8,16,20,24,26,28,32], 'itemsize':36}`

The mismatch in data layout will  fail an assertion.

The `dtype_from_fields` function does allow for the point_step (itemsize) to be specficied explicitly when there is padding present at the end of the data, but there is currently no way to pass this in from the  `create_cloud` function.

Fix adds an optional parameter to allow users to pass in a custom point_step, allowing the data to align properly.


clalancette (changes requested):
Can you please add a test showing that this PR fixes the problem you are describing?

broughtong:
Done :)

clalancette (approved):
This looks reasonable to me, thanks for the additional tests.  I'll run CI on it next.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18850/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13391/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19604/)

PR #231, opened 2023-11-06
sensor_msgs/CompressedImage: updated description of format field

ijnek:
Port of https://github.com/ros/common_msgs/pull/184 to ROS 2.

From original PR:

> 
> The description of the format field has long been outdated. This PR fixes it to be up-to-date with the core image_transport plugins.
>

### DIFF
--- a/sensor_msgs/msg/CompressedImage.msg
+++ b/sensor_msgs/msg/CompressedImage.msg
@@ -8,7 +8,34 @@ std_msgs/Header header # Header timestamp should be acquisition time of image
                              # +z should point into to plane of the image
 
 string format                # Specifies the format of the data
-                             #   Acceptable values:
-                             #     jpeg, png, tiff
+                             # Acceptable values differ by the image transport used:
+                             # - compressed_image_transport:
+                             #     ORIG_PIXFMT; CODEC compressed [COMPRESSED_PIXFMT]
+                             #   where:
+                             #   - ORIG_PIXFMT is pixel format of the raw image, i.e.
+                             #     the content of sensor_msgs/Image/encoding with
+                             #     values from include/sensor_msgs/image_encodings.h
+                             #   - CODEC is one of [jpeg, png, tiff]
+                             #   - COMPRESSED_PIXFMT is only appended for color images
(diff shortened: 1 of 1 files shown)


fujitatomoya (approved):
lgtm

fujitatomoya:
@mjcarroll this one comes from https://github.com/ros/common_msgs/pull/184, can you take a look?

clalancette (approved):
In short, this looks good to me and does seem to correspond to what the code currently supports.  I'm going to go ahead and run CI on it, thanks!

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19799/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14303/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20523/)

ijnek:
Thanks @fujitatomoya and @clalancette !

amacneil:
> The description of the format field has long been outdated. This PR fixes it to be up-to-date with the core image_transport plugins.

Is there any info on when this became outdated? Fwiw I'm pretty sure we support only the "simple" values and not the "verbose" ones in Foxglove, and I don't recall any complaints or feature requests about this, so I'm surprised to hear that the simple ones are unused.

> If the field is empty or does not correspond to the above pattern, the image is treated as png image.

If I'm reading this change correctly, the old "simple" format values of `jpeg` and `png` are now unsupported, and `jpeg` should be interpreted as `png`?

ijnek:
I don't have details on when simple jpeg, png, etc. were dropped (it may still work with compressed_image_transport, i'll have to give it a try), but comprssed publisher and compressed depth publisher from image_transport_plugins has been using the new format for at least 12 years (https://github.com/ros-perception/image_transport_plugins/commit/e9c27689dbe246e3bf23e097a4e38bcb0dfad9de). 

As a side note, since compressed images and compressed depth images both use `CompressedImage`, formats like "png" could be ambiguous since both compressed png image and compressed depth png image exist.

Issue #230, opened 2023-10-30
Adding unique identifer field to Polygon/Polygon Stamped


SteveMacenski:
As part of https://github.com/ros-planning/navigation2/pull/3885, we're interested in representing footprints as polygons and storing them to do collision monitoring with respect to for multi-robot fleets.

Managing the `N` polygons and new incoming messages is complicated by the fact that the standard Polygon message has no form of unique ID field (int, uuid, etc) to identify one physical object instance being represented at `t=T`.

I would like to propose adding to PolygonStamped and potentially Polygon a `uint16 id` field as http://docs.ros.org/en/noetic/api/visualization_msgs/html/msg/Marker.html and others have themselves for similar deconfliction of unique objects.

SteveMacenski:
It would look like

```
# This represents a Polygon with reference coordinate frame and timestamp

std_msgs/Header header
Polygon polygon
uint16 id
```

Which would add minimal overhead to the message definitions and not change any existing fields' name or type, so it should not break anyone's current use-cases

SteveMacenski:
@clalancette who's the right person to ping about this?

SteveMacenski:
https://github.com/ros2/common_interfaces/pull/232 implements

ros-discourse:
This issue has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2023-11-16/34757/1

PR #232, opened 2023-11-13
adding IDs to geometry_msgs/Polygon, PolygonStamped

SteveMacenski:
Implementing https://github.com/ros2/common_interfaces/issues/230

Polygons are often used to represent specific objects (e.g. robot footprints, obstacles in the scene, other agents) but are difficult to rectify currently without any kind of specific identification - especially when you have `N` of the same object resulting in several identical object instances. 

This PR adds an ID field analog to `visualization_msgs/Marker` to disambiguate polygons for use in the `nav2_collision_monitor` and throughout future `Costmap2D` planned work. I strongly believe this is a good addition worth the change to a standard interface with both minimal overhead & not breaking any current users' behavior.

### DIFF
--- a/geometry_msgs/msg/PolygonInstance.msg
+++ b/geometry_msgs/msg/PolygonInstance.msg
@@ -1,5 +1,5 @@
 # A specification of a polygon where the first and last points are assumed to be connected
 # Includes and unique identification field for disambiguating multiple instances
 
-Point32[] points
+geometry_msgs/Polygon polygon
 int64 id
--- a/geometry_msgs/msg/PolygonInstance.msg
+++ b/geometry_msgs/msg/PolygonInstance.msg
@@ -1,3 +1,5 @@
 # A specification of a polygon where the first and last points are assumed to be connected
+# Includes and unique identification field for disambiguating multiple instances
 
 Point32[] points
+int64 id
--- a/geometry_msgs/msg/PolygonInstance.msg
+++ b/geometry_msgs/msg/PolygonInstance.msg
@@ -1,5 +1,5 @@
 # A specification of a polygon where the first and last points are assumed to be connected
-# Includes and unique identification field for disambiguating multiple instances
+# It includes a unique identification field for disambiguating multiple instances
 
 geometry_msgs/Polygon polygon
 int64 id
--- a/geometry_msgs/msg/PolygonInstanceStamped.msg
+++ b/geometry_msgs/msg/PolygonInstanceStamped.msg
@@ -1,0 +1,6 @@
+# This represents a Polygon with reference coordinate frame and timestamp
+# Includes and unique identification field for disambiguating multiple instances
+
+std_msgs/Header header
+Point32[] points
+int64 id
--- a/geometry_msgs/msg/PolygonInstanceStamped.msg
+++ b/geometry_msgs/msg/PolygonInstanceStamped.msg
@@ -1,5 +1,5 @@
 # This represents a Polygon with reference coordinate frame and timestamp
-# Includes and unique identification field for disambiguating multiple instances
+# It includes a unique identification field for disambiguating multiple instances
 
 std_msgs/Header header
 geometry_msgs/PolygonInstance polygon
--- a/geometry_msgs/msg/PolygonInstanceStamped.msg
+++ b/geometry_msgs/msg/PolygonInstanceStamped.msg
@@ -1,5 +1,6 @@
 # This represents a Polygon with reference coordinate frame and timestamp
 # Includes and unique identification field for disambiguating multiple instances
 
-geometry_msgs/PolygonStamped polygon
+std_msgs/Header header
+geometry_msgs/Polygon polygon
 int64 id
--- a/geometry_msgs/msg/PolygonInstanceStamped.msg
+++ b/geometry_msgs/msg/PolygonInstanceStamped.msg
@@ -2,5 +2,4 @@
 # Includes and unique identification field for disambiguating multiple instances
 
 std_msgs/Header header
-geometry_msgs/Polygon polygon
-int64 id
+geometry_msgs/PolygonInstance polygon
--- a/geometry_msgs/msg/PolygonInstanceStamped.msg
+++ b/geometry_msgs/msg/PolygonInstanceStamped.msg
@@ -1,6 +1,5 @@
 # This represents a Polygon with reference coordinate frame and timestamp
 # Includes and unique identification field for disambiguating multiple instances
 
-std_msgs/Header header
-Point32[] points
+geometry_msgs/PolygonStamped polygon
 int64 id
--- a/geometry_msgs/CMakeLists.txt
+++ b/geometry_msgs/CMakeLists.txt
@@ -26,6 +26,8 @@ set(msg_files
   "msg/Point32.msg"
   "msg/PointStamped.msg"
   "msg/Polygon.msg"
+  "msg/PolygonInstance.msg"
+  "msg/PolygonInstanceStamped.msg"
   "msg/PolygonStamped.msg"
   "msg/Pose.msg"
   "msg/Pose2D.msg"
--- a/geometry_msgs/README.md
+++ b/geometry_msgs/README.md
@@ -15,6 +15,8 @@ For more information about ROS 2 interfaces, see [docs.ros.org](https://docs.ros
 * [Point](msg/Point.msg): The position of a 3-dimensional point in free space.
 * [PointStamped](msg/PointStamped.msg): Point with reference coordinate frame and timestamp.
 * [Polygon](msg/Polygon.msg): A specification of a polygon where the first and last points are assumed to be connected.
+* [PolygonInstance](msg/PolygonInstance.msg): A specification of a polygon where the first and last points are assumed to be connected. Contains an identification field for disambiguation of multiple instances.
+* [PolygonInstanceStamped](msg/PolygonInstanceStamped.msg): A Polygon with reference coordinate frame and timestamp. Contains an identification field for disambiguation of multiple instances.
 * [PolygonStamped](msg/PolygonStamped.msg): A Polygon with reference coordinate frame and timestamp.
 * [Pose2D](msg/Pose2D.msg): **Deprecated as of Foxy and will potentially be removed in any following release.**
 * [PoseArray](msg/PoseArray.msg): An array of poses with a header for global reference.
--- a/geometry_msgs/msg/Polygon.msg
+++ b/geometry_msgs/msg/Polygon.msg
@@ -1,3 +1,4 @@
 # A specification of a polygon where the first and last points are assumed to be connected
 
 Point32[] points
+int32 id
--- a/geometry_msgs/msg/PolygonStamped.msg
+++ b/geometry_msgs/msg/PolygonStamped.msg
@@ -2,4 +2,3 @@
 
 std_msgs/Header header
 Polygon polygon
-int32 id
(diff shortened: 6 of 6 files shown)


peci1:
Is this really needed? Wouldn't it be better to add a `PolygonCollision` to `nav_msgs`? I always understood the geometry_msgs as very general. E.g. I have a node that publishes the supporting polygon of the robot. What would the ID be in that case?

SteveMacenski:
Polygons are used to represent numerous things that having some kind of ID would be useful for.  

Examples:
- `N` homogenous robots in a room, representing each in the global frame wouldn't tell me which polygon belongs to what robot (collision use; non collision use; fleet use; ...)
- I have a bunch of fields I want to plow and/or regions I want to represent in a warehouse. I want to be able to know which field / zone this polygon belongs to correlate some metadata 
- I want to represent some virtual information like sensor frustum bounds of many instances of the same sensor in the global frame. 


I disagree that this should be some navigation specific message type or particular to collision. I think the general polygon message is missing key information to be able to distinguish between polygons sharing the same `frame_id`. Other messages in `geometry_msgs` have unique frame IDs or are not application-wise necessary to distinguish them. Polygons are more generic and do, I feel, require some kind of information to distinguish them from one another for metadata correlation reasons in many unique applications

peci1:
> * I have a bunch of fields I want to plow and/or regions I want to represent in a warehouse. I want to be able to know which field / zone this polygon belongs to correlate some metadata

Ehh, really should this be supported by a core message type? This sounds to me like a cartesian version of geographic_msgs.

> * Something something, detection box polygons from AI 😉

Sounds more like vision_msgs.

> * `N` homogenous robots in a room, representing each in the global frame wouldn't tell me which polygon belongs to what robot (collision use; non collision use; fleet use; ...)

> * I want to represent some virtual information like sensor frustum bounds of many instances of the same sensor in the global frame.

These two seem equivalent. Usually each robot/sensor will have its own topic or frame_id, so you could disambiguate based on that. If you're building a list part by part by listening for single elements on a topic, why couldn't the elements come in their original (unique) frame and you couldn't convert them inside the node?

> I think that is one of the motivating reasons why [visualization_msgs/Marker](https://github.com/ros2/common_interfaces/blob/rolling/visualization_msgs/msg/Marker.msg#L31) has an ID field

But Marker is a special-purpose message type, while Polygon is a general-purpose message type. Marker's purpose dictates that having an id is essential. You can't say that generally about polygons.

Regarding type of the id field:
* https://github.com/ros-perception/vision_msgs/issues/47 settled on string
* https://github.com/ros-perception/radar_msgs/pull/3 settled on uuid

Do you really want to introduce yet another id type? :)

---

All the thinking around leads me to proposing a new message type instead of editing the existing one. I'm not sure about the name, but something along `PolygonWithIdentity`, which would correspond to your proposed edit of `PolygonStamped`. A small helper library could be provided that would automatically take original `PolygonStamped` and add the id based e.g. on the topic name, or that would take the local polygon, transform it to global frame and assign an ID based on the local frame.

SteveMacenski:
> Usually each robot/sensor will have its own topic or frame_id, so you could disambiguate based on that.

Not when they're already in the global frame (or base frame, etc). That's the exact need here, to disambiguate with many instances of similar / same objects.

> I'm not sure about the name, but something along PolygonWithIdentity

I personally think that's overkill for what complexity we're talking about, but I wouldn't object to that. The biggest thing is that this is a generic problem that needs a generic `geometry_msgs` solution (I feel) so that it can be leveraged in many places. I don't think its a strict need that it is in precisely `Polygon` base message.

How about `PolygonInstance`  and `PolygonInstanceStamped`?

peci1:
> How about `PolygonInstance` and `PolygonInstanceStamped`?

These names sound good to me ;)

SteveMacenski:
Maintainers et al of this repository: Any objection to this course of action?

clalancette:
> Maintainers et al of this repository: Any objection to this course of action?

We discussed this at the triage meeting this week, and we think this is a fine course of action.  We do have a couple of comments and questions:

1.  It seems like the ID would be better as a `uint64`.  That way you could use e.g. nanosecond timestamps as the ID if you wished.  Another alternative would be a string (at which point you could put a full UUID in there), but that may be overkill for what you want.
2. Speaking of the ID, do you have a way to generate it in mind?  In other words, what are you expecting to put in that field?

SteveMacenski:
@clalancette `uint64` sounds good, I'll make that update when I make the new PolygonInstance and PolygonInstanceStamped messages. I'll get on that once its all settled, I suppose just your next question and we're good to go?

> do you have a way to generate it in mind? 

For any particular application, there's a way to do this. So if the polygons are robots, using a serial number seems reasonable. If this is the output of AI or some other thing, UUID generated from the tracking software seems reasonable. That to me is a potential argument for using a `string`, but I don't feel strongly about it. Correlating a UUID to a ID from a tracker isn't difficult. 

The difficulty would come if you had `N` publishers of these PolygonInstance messages that don't know about each other and have their own index-from-0 schema for generating IDs. When based on something physical or annotated, then the numbering schema from serial numbers or something makes sense, but if each publisher is generating virtual polygons, UUID is the most reasonable way to go to avoid conflicts. I don't think having this field as a UUID is required, but a string would do that job.

I leave that choice of `string` vs `uint64` to maintainers. I lean slightly towards `string` personally for the above reasons, but using `int` IDs has heritage in other messages and I do value consistency

clalancette:
> @clalancette `uint64` sounds good, I'll make that update when I make the new PolygonInstance and PolygonInstanceStamped messages.

I think uint64 is good enough; if we need the additional resolution that an e.g. string would give us, we can get that in different ways (separate topics, a higher-level message that includes this message + robot_id, etc).

SteveMacenski:
What is a 'standard message' but a message that can be used to describe common and innumerable-situation things for many applications. I think a way to identify a shape against other shapes in the same global frame has enough independent applications that this is worthwhile. Especially when many different types of representations that can be generalized to this (agent, obstacle, footprint) have some relationship that some applications may want to ingest all of them and don't need to specifically differentiate between the semantics of each of those 3 example types, as much as it just needs to be able to disambiguate between instances as updates are made. 

I think strongly that this is very relevant and I have ~3 places within Nav2 and broader tooling around it that this would get use within a couple of months and one that would be used immediately.

peci1:
> It sounds like there's support for adding the ID to the message. I disagree a little about that. It seems like an id isn't needed to define a polygon. I would instead recommend a higher level message type for a robot footprint, or an obstacle, or an agent such that it has both an id and a polygon.

I thought the consensus was to add a new PolygonInstance message type.

SteveMacenski:
It is/was.

@clalancette @peci1 can you rereview? This is blocking a user https://github.com/ros-planning/navigation2/pull/3885

peci1 (approved):

SteveMacenski:
Squash merge should handle :-) 

Great!

geometry_msgs/msg/PolygonInstance.msg
```diff
@@ -0,0 +1,5 @@
+# A specification of a polygon where the first and last points are assumed to be connected
+# Includes and unique identification field for disambiguating multiple instances
+
+Point32[] points
```
clalancette, line 4: Looking at this in Waffle, we think that this should be a `Polygon`; that way, APIs that take a `geometry_msgs::msg::Polygon` can still work with this.

geometry_msgs/msg/PolygonInstanceStamped.msg
```diff
@@ -0,0 +1,6 @@
+# This represents a Polygon with reference coordinate frame and timestamp
+# Includes and unique identification field for disambiguating multiple instances
+
+std_msgs/Header header
+Point32[] points
+int64 id
```
clalancette, line 6: Similarly, to conform to our usual practice this should be a `PolygonInstance`.

clalancette (changes requested):

SteveMacenski:
Done!

geometry_msgs/msg/PolygonInstanceStamped.msg
```diff
@@ -0,0 +1,5 @@
+# This represents a Polygon with reference coordinate frame and timestamp
+# Includes and unique identification field for disambiguating multiple instances
+
+geometry_msgs/PolygonStamped polygon
```
peci1, line 4: Not sure how this works in ROS 2, but in ROS 1 there is an explicit requirement to have header as the first field to qualify for the special handling of stamped messages.

mjcarroll: I believe it is using the same logic in ROS 2, I would expect:

```
std_msgs/Header header
geometry_msgs/PolygonInstance polygon
```

SteveMacenski: Agreed, change made

peci1 (approved):

SteveMacenski:
Failure looks like networking related, not PR related

geometry_msgs/msg/PolygonInstanceStamped.msg
```diff
@@ -0,0 +1,6 @@
+# This represents a Polygon with reference coordinate frame and timestamp
+# Includes and unique identification field for disambiguating multiple instances
+
+std_msgs/Header header
+geometry_msgs/Polygon polygon
+int64 id
```
mjcarroll, line 6: Sorry, one more quibble, this would make it more in-line with the other `FooStamped` messages.

```suggestion
std_msgs/Header header
geometry_msgs/PolygonInstance polygon
```

SteveMacenski:
Quibble away, my friend :wink:

mjcarroll (approved):

mjcarroll:
I'm happy with this, it has been sufficiently bike-shedded.

sloretz (approved):

SteveMacenski:
Got a few +1’s, can we merge this?

geometry_msgs/msg/PolygonInstance.msg
```diff
@@ -0,0 +1,5 @@
+# A specification of a polygon where the first and last points are assumed to be connected
+# Includes and unique identification field for disambiguating multiple instances
```
clalancette, line 2:
```suggestion
# It includes a unique identification field for disambiguating multiple instances
```

geometry_msgs/msg/PolygonInstanceStamped.msg
```diff
@@ -0,0 +1,5 @@
+# This represents a Polygon with reference coordinate frame and timestamp
+# Includes and unique identification field for disambiguating multiple instances
```
clalancette, line 2:
```suggestion
# It includes a unique identification field for disambiguating multiple instances
```

clalancette (changes requested):
Two really small changes to the wording, then I'll go ahead and run CI on this.

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20095/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14542/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20815/)

SteveMacenski:
I dont believe the windows linting is from this

clalancette:
> I dont believe the windows linting is from this

No, it's not.  It timed out in rclcpp, which is likely due to https://github.com/ros2/rclcpp/pull/2303 .  We'll deal with that separately.

ros-discourse:

https://discourse.ros.org/t/ros-2-jazzy-jalisco-released/37862/1

sloretz:
@mergifyio backport humble

PR #233, opened 2023-12-01
Create new messages with all fields needed to define a velocity and transform it

destogl:
This PR proposes a new message type for a full description of velocities, unambiguous definition, and the possibility to transform velocities using TF (this is currently commented out in `geometry2` library).

As found in the literature and comments across the `geometry2` library, `geometry_msgs/TwistStamped` is insufficient information for the complete representation.

The new fields should be sufficiently described in the message. If not clear, please provide feedback for improvement.

@tfoote @SteveMacenski @azeey  looking forward to your comments.

References:
- See commit ea03b20802e4eff5ae339f6a9042ceef51dd6665 in this repository from 24th Apr. 2010
- https://core.ac.uk/download/pdf/154240607.pdf
- http://hades.mech.northwestern.edu/images/7/7f/MR.pdf Eq'n 3.83


geometry_msgs/msg/VelocityStamped
```diff
@@ -0,0 +1,10 @@
+# This expresses the timestamped velocity of a frame 'body_frame_id' in the reference frame 'header.frame_id' expressed from arbitrary observation pose 'observation_pose'.
+# The message is often used in simplified versions where:
+# - observation pose is not defined, i.e., it is assumed that the observation pose is the origin of the 'header.frame_id';
+# - 'body_frame_id' and 'header.frame_id' are identical, i.e., the velocity is observed and defined in the local coordinates system of the body 
+#   which is the usual use-case in mobile robotics.
```
SteveMacenski, line 5: I think you can remove this particular line

geometry_msgs/msg/VelocityStamped
```diff
@@ -0,0 +1,10 @@
+# This expresses the timestamped velocity of a frame 'body_frame_id' in the reference frame 'header.frame_id' expressed from arbitrary observation pose 'observation_pose'.
+# The message is often used in simplified versions where:
+# - observation pose is not defined, i.e., it is assumed that the observation pose is the origin of the 'header.frame_id';
```
SteveMacenski, line 3: I'd add a bullet: If the observation pose is an established frame, you may set its frame_id only in the pose (since the relative pose to that frame is Identity). It would be good to make folks note how to use this if everything at hand is a TF frame

SteveMacenski (changes requested):

geometry_msgs/msg/VelocityStamped
destogl:
```suggestion
# - observation pose is not defined, i.e., it is assumed that the observation pose is the origin of the 'header.frame_id';
# - if the observation pose is an established frame, you may set its frame_id only in the pose (since the relative pose to that frame is Identity);
```

geometry_msgs/msg/VelocityStamped
destogl: I am with you on this, my only concern is not to confuse people that are moving from Twist as used currently in Nav2 and ros2_control.

destogl:
```suggestion
#   as common in mobile robotics.
```

destogl:
@tfoote any comments on this?

tfoote (approved):
The changes look good. I'd like to flesh out the documentation more completely so that people in the future clearly understand the design. As well as confirming that we can confirm that the transform velocity operations have enough semantic information to be restored.

destogl:
Great, feel free to add any comments. It would be great if we can merge this sometime in February, so we can adjust controllers already for Jazzy.

geometry_msgs/msg/VelocityStamped
azeey: I'm a little confused. if `body_frame_id` and `header.frame_id` are identical, shouldn't the velocity be always zero? i.e, we're asking the velocity of a body relative to itself. On the other hand, we could have `body_frame_id` and `observation_pose` (via its `frame_id`) be identical and `header.frame_id`  be something different from the two. This makes sense to me since it's asking the velocity of the body relative to whatever `header.frame_id` is but expressed in the body frame.

audrow:
@tfoote and @destogl, it sounds like this is waiting on documentation. What's the status of this? Also, any thoughts on what exactly should be done before this is ready to go in?

ahcorde (changes requested):
I will be working int this feature https://github.com/ros2/geometry2/issues/643 but this PR requires some changes to be operational

 - Modify name `geometry_msgs/msg/VelocityStamped` to `geometry_msgs/msg/VelocityStamped.msg`
 - Include this new message in the `CMakeLists.txt`

@destogl do you think you can include these changes?

tfoote:
I think that this is very close, but before merging I'd like to make sure that we have both documentation but also the example implementation that validates that we have actually covered all our bases.

ahcorde:
I added this PR https://github.com/StoglRobotics-forks/common_interfaces/pull/1 to include the required changes

destogl:
@tfoote what do you mean about documentation? Should be adapt this in the message "header" or do you have something else in mind?Š

tfoote:
The names right now are not referencing a standard of notation and talking to others, they have made different assumptions about what each field means. This is coming to a head as we're trying to implement the usage in geometry2. I'm going to try to make a descriptive doc we can use as a baseline. But this message needs to be documented in the comments enough that someone can use that self documenting feature to interpret the data unambiguously.

destogl:
OK, can you make a proposal?

ahcorde:
Ey @destogl, @azeey and @SteveMacenski , 

@tfoote added this documentation to detail things https://github.com/ros2/ros2_documentation/pull/4235 

Feel free to add any comments

geometry_msgs/msg/VelocityStamped.msg
```diff
@@ -0,0 +1,11 @@
+# This expresses the timestamped velocity of a frame 'body_frame_id' in the reference frame 'header.frame_id' expressed from arbitrary observation pose 'observation_pose'.
+# The message is often used in simplified versions where:
+# - observation pose is not defined, i.e., it is assumed that the observation pose is the origin of the 'header.frame_id';
+# - if the observation pose is an established frame, you may set its frame_id only in the pose (since the relative pose to that frame is Identity);
+# - 'body_frame_id' and 'header.frame_id' are identical, i.e., the velocity is observed and defined in the local coordinates system of the body 
+#   which is the usual use-case in mobile robotics.
+
+std_msgs/Header header
+string body_frame_id
+PoseStamped observation_pose
```
tfoote, line 10:
```suggestion
string space_frame_id
```

geometry_msgs/msg/VelocityStamped.msg
```diff
@@ -0,0 +1,11 @@
+# This expresses the timestamped velocity of a frame 'body_frame_id' in the reference frame 'header.frame_id' expressed from arbitrary observation pose 'observation_pose'.
```
tfoote, line 1:
```suggestion
# This expresses the velocity of a frame 'body_frame_id' in the space frame 'space_frame_id' expressed from arbitrary observation pose 'header.frame_id' at time `header.stamp`.
```

geometry_msgs/msg/VelocityStamped.msg
```diff
@@ -0,0 +1,11 @@
...
+# The message is often used in simplified versions where:
+# - observation pose is not defined, i.e., it is assumed that the observation pose is the origin of the 'header.frame_id';
+# - if the observation pose is an established frame, you may set its frame_id only in the pose (since the relative pose to that frame is Identity);
+# - 'body_frame_id' and 'header.frame_id' are identical, i.e., the velocity is observed and defined in the local coordinates system of the body 
+#   which is the usual use-case in mobile robotics.
```
tfoote, line 6:
```suggestion
#   which is the usual use-case in mobile robotics and is also known as a body twist.
```

geometry_msgs/msg/VelocityStamped.msg
```diff
@@ -0,0 +1,11 @@
+# This expresses the timestamped velocity of a frame 'body_frame_id' in the reference frame 'header.frame_id' expressed from arbitrary observation pose 'observation_pose'.
+# The message is often used in simplified versions where:
+# - observation pose is not defined, i.e., it is assumed that the observation pose is the origin of the 'header.frame_id';
+# - if the observation pose is an established frame, you may set its frame_id only in the pose (since the relative pose to that frame is Identity);
+# - 'body_frame_id' and 'header.frame_id' are identical, i.e., the velocity is observed and defined in the local coordinates system of the body
```
tfoote, line 5:
```suggestion
# - If the 'body_frame_id' and 'header.frame_id' are identical, i.e., the velocity is observed and defined in the local coordinates system of the body 
```

geometry_msgs/msg/VelocityStamped.msg
```diff
@@ -0,0 +1,11 @@
+# This expresses the timestamped velocity of a frame 'body_frame_id' in the reference frame 'header.frame_id' expressed from arbitrary observation pose 'observation_pose'.
+# The message is often used in simplified versions where:
+# - observation pose is not defined, i.e., it is assumed that the observation pose is the origin of the 'header.frame_id';
```
tfoote, line 3:
```suggestion
# - If the 'space_frame_id' and 'header.frame_id' are identical, i.e., the velocity is observed and defined in the local coordinates system of the spacial frame.  
```

geometry_msgs/msg/VelocityStamped.msg
```diff
@@ -0,0 +1,11 @@
...
+# - if the observation pose is an established frame, you may set its frame_id only in the pose (since the relative pose to that frame is Identity);
```
tfoote, line 4:
```suggestion
#   This is also known as a spacial twist.
```

tfoote (commented):
Here's my concrete suggestions for changes to follow those external conventions. I think that following those as closely as possible will hopefully decrease the amount of confusion in the long run.

tfoote:
Thanks to @azeey for following up and flagging that this isn't actually spacial twist. Using the reference https://www.cse.lehigh.edu/~trink/Courses/RoboticsII/reading/murray-li-sastry-94-complete.pdf Pages 54-55 
"A Mathematical Introduction to
Robotic Manipulation
Richard M. Murray
California Institute of Technology
Zexiang Li
Hong Kong University of Science and Technology
S. Shankar Sastry
University of California, Berkeley
"

So I will update my suggestions based on that. 

When the frame in which the twist is represented is the same as the reference frame then what we're representing the body twist offset to that frame and not changing the math to be a spacial twist. 

And to that end we're simply transforming the body twist to the new coordinate frame. 

So to pull this all back together we will represent a body twist between `body_frame_id` and `reference_frame_id` and then transform it to be represented in the observational frame identified by `header.frame_id` I'll update my changes accordingly.

geometry_msgs/msg/VelocityStamped.msg
tfoote:
```suggestion
string reference_frame_id
```

Changing this to reference as this is the frame of reference to which the velocity is being measured.

geometry_msgs/msg/VelocityStamped.msg
tfoote:
```suggestion
# This expresses the timestamped velocity vector of a frame 'body_frame_id' in the reference frame 'reference_frame_id'  transformed to be expressed from arbitrary observation frame 'header.frame_id'.
```

geometry_msgs/msg/VelocityStamped.msg
```diff
@@ -0,0 +1,11 @@
+# This expresses the timestamped velocity of a frame 'body_frame_id' in the reference frame 'header.frame_id' expressed from arbitrary observation pose 'observation_pose'.
+# The message is often used in simplified versions where:
```
tfoote, line 2:
```suggestion
```

With the removal of the PoseStamped and only using the frame_ids this optional version is no longer necessary.

geometry_msgs/msg/VelocityStamped.msg
tfoote, line 3:
```suggestion
```

Also removed

geometry_msgs/msg/VelocityStamped.msg
tfoote, line 4:
```suggestion
```

Also removed

ahcorde:
I created a new PR https://github.com/ros2/common_interfaces/pull/240 to include the suggestions that Tully made

tfoote:
#240 had been merged closing this one. 5.3.3 is available in rolling testing now with these changes https://github.com/ros/rosdistro/pull/40550

PR #235, opened 2023-12-13
ignore local project files

mosfet80:
exclude from push local project files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,7 @@
 # Python temporary files
 *.pyc
 __pycache__/
+
+# Local project files 
+*.cproject
+*.project


fujitatomoya (commented):
i am assuming that those files are specific to Eclipse IDE?

if that is right, how about adding your own `.gitignore` to your git configuration?

- so that we(ros2 all repos) do not need to maintain many IDE specific files to `.gitignore`.
- this seems to be user configuration, but repository. so adding `.gitignore` to user environment should work much better than this approach? e.g I do use emacs/ctags/global/cscope, those specific `tag` files are set to `.gitignore`, and it works for me.

thanks,

mosfet80:
> i am assuming that those files are specific to Eclipse IDE?
> 
> if that is right, how about adding your own `.gitignore` to your git configuration?
> 
> * so that we(ros2 all repos) do not need to maintain many IDE specific files to `.gitignore`.
> * this seems to be user configuration, but repository. so adding `.gitignore` to user environment should work much better than this approach? e.g I do use emacs/ctags/global/cscope, those specific `tag` files are set to `.gitignore`, and it works for me.
> 
> thanks,

PR #238, opened 2024-02-28
Use target qualifier for checking the cpp typesupport exists

Ryanf55:
Follow up from https://github.com/ros-geographic-info/geographic_info/pull/56#discussion_r1505260952, where I recommend we check this condition based on the `TARGET` keyword. We are expecting a target, not a non-zero length string, so it better expresses intent.

I would like to set the precedent for it here, and then perhaps update the ROS tutorial on this feature.

### DIFF
--- a/sensor_msgs/CMakeLists.txt
+++ b/sensor_msgs/CMakeLists.txt
@@ -64,7 +64,7 @@ if(BUILD_TESTING)
   ament_lint_auto_find_test_dependencies()
 endif()
 
-if(cpp_typesupport_target)
+if(TARGET "${cpp_typesupport_target}")
   add_library(${PROJECT_NAME}_library INTERFACE)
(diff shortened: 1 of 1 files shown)


sloretz (approved):

sloretz:
CI ([repos](https://gist.githubusercontent.com/sloretz/175b6642adb6bed073e13e414c95a6ae/raw/a117813e4626571e37db880517d1cfa5f6973390/ros2.repos) build: `--packages-up-to sensor_msgs` test: `--packages-select sensor_msgs`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20438/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14840/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21181/)

sloretz:
Thank you for the PR!

PR #242, opened 2024-04-10
Adds ARROW_STRIP to Marker.msg

tfoote:
This is a rebase of #218 for a clean merge from @rr-tom-noble but since it's been queued so long it's out of date. 

There's a partial implementation in https://github.com/ros2/rviz/pull/972

However I would suggest that we land this in rolling to both keep it in sync with the ROS 1 version that's already added this. And to enable the use of the message with other implementations in the future. If we don't add this now it won't be able to be added for another year.

### DIFF
--- a/visualization_msgs/msg/Marker.msg
+++ b/visualization_msgs/msg/Marker.msg
@@ -16,6 +16,7 @@ int32 POINTS=8
 int32 TEXT_VIEW_FACING=9
 int32 MESH_RESOURCE=10
 int32 TRIANGLE_LIST=11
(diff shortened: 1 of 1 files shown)


ahcorde (approved):

tfoote:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20822/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15126/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21509/)

PR #243, opened 2024-04-12
Removed TODO

ahcorde:
There is a TODO that mentioned the adition of this message. Happy to remove the comment and the new message if this is not required anymore

There is already a service called GetMap.srv

@SteveMacenski, @tfoote  ?

### DIFF
--- a/nav_msgs/CMakeLists.txt
+++ b/nav_msgs/CMakeLists.txt
@@ -17,6 +17,10 @@ find_package(geometry_msgs REQUIRED)
 find_package(rosidl_default_generators REQUIRED)
 find_package(std_msgs REQUIRED)
 
+set(action_files
+  "action/GetMap.action"
+)
+
 set(msg_files
   "msg/GridCells.msg"
   "msg/MapMetaData.msg"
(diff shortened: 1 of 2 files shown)


SteveMacenski:
This seems like a weird action, I think a service is the most appropriate interface - but I suppose it doesn't _hurt_. Has anyone been requesting this?

ahcorde:
Nope, I just saw the TODO

SteveMacenski:
My feedback is "lets just remove the TODO and see if anyone cares" since it doesn't seem that anyone has until now. But, hey, on the other hand, its not hurting anyone.

clalancette:
> My feedback is "lets just remove the TODO and see if anyone cares" since it doesn't seem that anyone has until now. But, hey, on the other hand, its not hurting anyone to add the new action.

I'd be more for removing the TODO.  We shouldn't add things in that don't have users, in my opinion.

tfoote:
Thanks for finding the TODO and reviewing it. But I think that the GetMap action doesn't add a lot of value on top of the GetMap Service so as there's no demand we might as well not add it and keep things simpler.

tfoote (changes requested):
Without feedback this doesn't add anything on top of the service. And we now have async service clients so that's not a benefit anymore either.

tfoote (approved):
Thanks for cleaning this up!

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20858/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15161/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21544/)

PR #244, opened 2024-04-16
Remove references to index.ros.org.

clalancette:


tfoote (approved):
We need to resolve referencing docs.ros.org vs index.ros.org overall and have a better flow between them. The shorter url and simpler pattern and multi version headers were one of the things that made the wiki very powerful. 

This embedding the rolling into the url is suboptimal. Maybe what we should be doing like the wiki is creating an autogenerated header with the index content on docs.ros.org that's embedded instead of on a standalone page, that can help with the cross referencing and navigation more easily.

clalancette:
> This embedding the rolling into the url is suboptimal.

Yeah, totally agreed.

> Maybe what we should be doing like the wiki is creating an autogenerated header with the index content on docs.ros.org that's embedded instead of on a standalone page, that can help with the cross referencing and navigation more easily.

Yeah, that is a good idea.  I still think we should take this for now, as at least we'll be pointing to a page that exists (even though it will be the incorrect version for e.g. Jazzy).

wjwwood (approved):

PR #257, opened 2024-09-08
reapply yuv encoding deprecation

christianrauch:
This reapplies https://github.com/ros2/common_interfaces/pull/247 with fixes for MSVC.

### DIFF
--- a/sensor_msgs/include/sensor_msgs/image_encodings.hpp
+++ b/sensor_msgs/include/sensor_msgs/image_encodings.hpp
@@ -122,6 +122,8 @@ const std::regex cv_type_regex("(8|16|32|64)(U|S|F)C([0-9]*)");
 // Utility functions for inspecting an encoding string
 static inline bool isColor(const std::string & encoding)
 {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
   return encoding == RGB8 || encoding == BGR8 ||
          encoding == RGBA8 || encoding == BGRA8 ||
          encoding == RGB16 || encoding == BGR16 ||
@@ -129,6 +131,7 @@ static inline bool isColor(const std::string & encoding)
          encoding == YUV422 || encoding == YUV422_YUY2 ||
          encoding == UYVY || encoding == YUYV ||
          encoding == NV21 || encoding == NV24;
+#pragma GCC diagnostic pop
 }
 
 static inline bool isMono(const std::string & encoding)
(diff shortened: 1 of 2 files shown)


clalancette:
I'm going to be perfectly frank; I'm not sure that this is worth putting in.  The simple fact is that these encodings are *not* standardized anywhere.  And I am fine having two aliases for the same thing.  Finally, this PR adds in a lot of `#ifdef` cruft.  The combination of all of these things says to me that this is making work for not a lot of gain.

That said, this is just my opinion, and I'm not going to fight hard for it.  If others think that this is much clearer, then I'm not going to put up a fuss about it.

christianrauch:
@ahcorde With https://github.com/ros2/rviz/pull/1276 now merged, can you run the CI again?

sensor_msgs/include/sensor_msgs/image_encodings.hpp
```diff
@@ -120,13 +122,25 @@ const std::regex cv_type_regex("(8|16|32|64)(U|S|F)C([0-9]*)");
 // Utility functions for inspecting an encoding string
 static inline bool isColor(const std::string & encoding)
 {
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable : 4996)
+#else
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
```
mjcarroll, line 131: For the sake of maintenance, can we scope down where the ifdef/pragmas are required?

I think we could use a function like:

```
static inline bool isPackedYUV(const std::string& encoding)
```

That contains the deprecation suppression, and then use that in the other functions?

christianrauch: Do you also want to use `isPackedYUV` as a public function? Eventually, the deprecated encodings will be removed. If this is a functionality that is supposed to be exposed to a user and used often, then I can add a `isPackedYUV`. But doing this just for the more narrow scoping of the encodings is a maintenance overhead IMHO.

christianrauch:
@ahcorde Can you run the CI again?

ahcorde:
Pulls: ros2/common_interfaces#257
Gist: https://gist.githubusercontent.com/ahcorde/3c3c497d78d15771820e5f1fe05e4d21/raw/ed0296b88c82fce76a31bf4360771600ef5803e1/ros2.repos
BUILD args:   --packages-above-and-dependencies common_interfaces --packages-above-and-dependencies common_interfaces
TEST args:  --packages-above common_interfaces --packages-above common_interfaces
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14885
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22197/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16449/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1718/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22947/)

christianrauch:
@tfoote @mjcarroll @ahcorde Can you have a look at this PR again, please? I had to rebase, but it should still pass all the OS configurations.

PR #258, opened 2024-10-03
Add StringArray

ufrhaidar:
useful sometimes.

Feels dumb to do that in my workspace, so hide it in common_interfaces.

### DIFF
--- a/std_msgs/CMakeLists.txt
+++ b/std_msgs/CMakeLists.txt
@@ -38,6 +38,7 @@ set(msg_files
   "msg/MultiArrayDimension.msg"
   "msg/MultiArrayLayout.msg"
   "msg/String.msg"
+  "msg/StringArray.msg"
   "msg/UInt16.msg"
   "msg/UInt16MultiArray.msg"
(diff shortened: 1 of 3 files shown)


fujitatomoya (commented):
i thought this sounds reasonable at 1st, but `std_msgs` does not have any array typed messages at all but `MultiArray`. i think this is intention to keep it in that way? 

@tfoote @clalancette what do you think?

ufrhaidar:
I originally just wanted to add an array of StringArrays to a custom message. Was hoping someone would suggest a better way to achieve that like `string[][]` 😂

Looks like it can be added to example_msgs like the rest.

fujitatomoya:
We had a discussion in PMC meeting. This is probably not gonna be taken, we try to keep the primitive types only. thank you very much for opening PR and idea.

ufrhaidar:
any reconsideration after #262 ?

tfoote:
That one has semantic information beyond an array of primatives. This does not. I'm asking for a rereview of that one before it's released though.

PR #262, opened 2024-12-03
Add PoseStampedArray

tonynajjar:
The need originated from [here](https://github.com/ros-navigation/navigation2/pull/4771#:~:text=Also%2C%20food%20for%20thought%2C%20this%20seems%20like%20a%20good%20geometry_msgs%20improvement%20rather%20than%20a%20nav2_msgs%20improvement%20to%20add%20PosesStamped.%20Maybe%20we%20should%20file%20a%20ticket%20over%20there%20and%20ask%20%40tfoote%20if%20he%27s%20open%20to%20that%20contribution). @SteveMacenski


mjcarroll (approved):
No arguments from me.

Should we potentially add a comment here about what the difference between the root `std_msgs/Header` and the per-message `std_msgs/Header` mean here, or is that entirely application specific.

SteveMacenski:
I think the global message stamp could be removed?

tonynajjar:
> I think the global message stamp could be removed?

Would you then also remove the header from this one? -> https://github.com/ros2/common_interfaces/blob/rolling/nav_msgs/msg/Path.msg
Same concept no?

SteveMacenski:
The path's header is for the path itself, the pose headers are for each of the poses in the path if it were a time denoted trajectory. Admittedly its rarely (if ever) used that way, but it is conceptually the right way for the path to be structured for kinodynamic paths, should uses need to represent one.

Does the application of `PoseStampedArray` have a similar need? I'm ambivalent, just couldn't think of a reason why we needed it for `PoseStampedArray`. I think to Michael's point, there's no objections, but if included we should explain how it should be used so putting that reason inline would resolve his concern and provide clarity. If we can't think of a reason, then maybe just remove it

tonynajjar:
Okay. I also don't see a reason to keep it but I wanted to stay consistent. Your explanation as to why `nav_msgs/Path` would have a header makes sense. I removed the header

tonynajjar:
While working with this message we actually found a potential use for the global header. I added a comment but see [here](https://github.com/ros-navigation/navigation2/pull/4771/files#r1868577905) for more info. Feel free to propose a reformulation

SteveMacenski (approved):

SteveMacenski:
@mjcarroll I think this is in the final form with the header description. Do we need another maintainer to review or can we ship?

geometry_msgs/msg/PoseStampedArray.msg
```diff
@@ -0,0 +1,5 @@
+# An array of stamped poses with a header
+# The global header will store the time at which the poses were computed (not to confuse with the stamps of the poses themselves)
```
clalancette:
```suggestion
# The global header will store the time at which the poses were computed (not to be confused with the stamps of the poses themselves)
```

geometry_msgs/msg/PoseStampedArray.msg
```diff
@@ -0,0 +1,5 @@
...
+
+std_msgs/Header header
+PoseStamped[] poses
```
clalancette: Please add a newline at the end of the file.

clalancette (changes requested):

tonynajjar:
Thanks for the review @clalancette , changes were made, anything else?

SteveMacenski:
We're also planning on using this for the NavigateToPose interface - which makes alot more sense than our `vector<PoseStamped>` for a list of goal representation so we can have a stamp to gauge the age of the original request. We're finding more places to use this by the day :-)

tonynajjar:
> We're also planning on using this for the NavigateToPose interface

NavigateThroughPoses*

tonynajjar:
Happy new year! (ping in disguise @clalancette 😄 )

mjcarroll (approved):

mjcarroll:
Pulls: ros2/common_interfaces#262
Gist: https://gist.githubusercontent.com/mjcarroll/e51db32e9e99205be4fa32319946c28b/raw/5e9a971bdd5ed5eb905513bd6477c1a698030d59/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15039
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22355/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16618/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1886/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23146/)

SteveMacenski:
@mjcarroll they've all passed :-)

ahcorde (approved):

gavanderhoorn:
It's surprising to see this merged, especially after the many discussions about extending `geometry_msgs` with more messages with low semantic value (of which the linked https://github.com/ros/common_msgs/issues/144 is an example).

What changed? Different maintainers, changed opinions, evolution of best-practices? All of the above?

Swepz:
Friendly reminder to bump tag :)
image

tfoote:
I'd like to review the naming and documentation of this before we make a release to be sure we've locked down the semantics of this message. I've added it to the PMC meeting agenda for next week.

PR #278, opened 2025-04-30
[shape_msgs] New Bounds Msg

DLu:
This is the definition of Bounds from https://github.com/ros-simulation/simulation_interfaces/pull/1

I feel it should be moved to this upstream package for some future distro.

### DIFF
--- a/shape_msgs/msg/Bounds.msg
+++ b/shape_msgs/msg/Bounds.msg
@@ -1,0 +1,25 @@
+# Bounds which are useful in several contexts, e.g. to avoid spawning with other object overlap,
+# or parking in a spot that is too small.
+# Certain goals or points might be valid for a small object, but not suitable for a large one,
+# or a differently shaped one.
+# Bounds can be also checked to ensure certain scenario conditions are met.
+# For entities, these limits are relative to entity's canonical link transform, following ROS rep-103 convention.
+
+# As bounds are optional in most interfaces, TYPE_EMPTY signifies empty bounds, to be understood as "unbounded".
+# Otherwise, the fields are expected to define a valid volume.
+# For spawning with a named pose, you should check whether the entity simulation model fits within the bounds
+# before calling SpawnEntity, to avoid overlaps and unstable behavior.
+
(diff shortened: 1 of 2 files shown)


mjcarroll:
Added to PMC agenda for May 6th.

shape_msgs/msg/Bounds.msg
```diff
@@ -0,0 +1,25 @@
...
+# For entities, these limits are relative to entity's canonical link transform, following ROS rep-103 convention.
+
+# As bounds are optional in most interfaces, TYPE_EMPTY signifies empty bounds, to be understood as "unbounded".
+# Otherwise, the fields are expected to define a valid volume.
+# For spawning with a named pose, you should check whether the entity simulation model fits within the bounds
+# before calling SpawnEntity, to avoid overlaps and unstable behavior.
+
+# bounds type
+uint8 TYPE_EMPTY         = 0      # No bounds. The points vector will be empty.
+uint8 TYPE_BOX           = 1      # Axis-aligned bounding box, points field should have two values,
```
sloretz, line 15: What frame is the axis aligned bounding box aligned to?

ros-discourse:

https://discourse.ros.org/t/ros-pmc-minutes-for-may-6-2025/43576/1

shape_msgs/msg/Bounds.msg
DLu: That's an entirely reasonable question, and looking back, there is no frame specification anywhere. @adamdbrw was there a conversation about this in https://github.com/ros-simulation/simulation_interfaces/pull/1

adamdbrw: It is specified:
For entities, these limits are relative to entity's canonical link transform, following ROS rep-103 convention.

tfoote:
There does not seem to be a significant difference between this proposal and https://github.com/ros2/common_interfaces/blob/rolling/shape_msgs/msg/SolidPrimitive.msg

What's the semantic difference and why does it have the slightly different list of primatives selected?

PR #279, opened 2025-04-30
Result.msg for Service Responses

DLu:
This is the definition of `Result.msg` as it is defined in https://github.com/ros-simulation/simulation_interfaces/pull/1

From the README:
> The standard includes a generic message for handling service responses, [Result.msg](https://github.com/ros-simulation/simulation_interfaces/blob/main/msg/Result.msg), which will likely not be included in the final version of the standard. Since it is generic, it will either be promoted to a common message or included in the service mechanism itself.

This PR is for discussing its promotion to `common_interfaces` in `std_msgs` in some future distro.

The other option is it gets integrated into the service mechanism itself, as it was in ROS 1.

### DIFF
--- a/std_msgs/msg/Result.msg
+++ b/std_msgs/msg/Result.msg
@@ -1,0 +1,18 @@
+# Result code and message for service calls.
+# Note that additional results for specific services can defined within them using values above 100.
+
+uint8 RESULT_FEATURE_UNSUPPORTED   = 0    # Feature is not supported by the simulator, check GetSimulatorFeatures.
+                                          # While feature support can sometimes be deduced from presence of corresponding
+                                          # service / action interface, in other cases it is about supporting certain
+                                          # call parameters, formats and options within interface call.
+uint8 RESULT_OK                    = 1
+uint8 RESULT_NOT_FOUND             = 2    # No match for input (such as when entity name does not exist).
+uint8 RESULT_INCORRECT_STATE       = 3    # Simulator is in an incorrect state for this interface call, e.g. a service
+                                          # requires paused state but the simulator is not paused.
+uint8 RESULT_OPERATION_FAILED      = 4    # Request could not be completed successfully even though feature is supported
+                                          # and the input is correct; check error_message for details.
+                                          # Implementation rule: check extended codes for called service
+                                          #  (e.g. SpawnEntity) to provide a return code which is more specific.
+
+uint8 result                              # Result to be checked on return from service interface call
+string error_message                      # Additional error description when useful.
--- a/std_msgs/msg/Result.msg
+++ b/std_msgs/msg/Result.msg
@@ -14,5 +14,5 @@ uint8 RESULT_OPERATION_FAILED      = 4    # Request could not be completed succe
                                           # Implementation rule: check extended codes for called service
                                           #  (e.g. SpawnEntity) to provide a return code which is more specific.
 
-uint8 result                              # Result to be checked on return from service interface call
+int32 result                              # Result to be checked on return from service interface call
 string error_message                      # Additional error description when useful.
--- a/std_msgs/msg/Result.msg
+++ b/std_msgs/msg/Result.msg
@@ -1,18 +1,41 @@
 # Result code and message for service calls.
-# Note that additional results for specific services can defined within them using values above 100.
+# Note that additional codes for specific services can defined within the service definition.
+# For example, you could define a service CalculateSquareRoot.srv as
+# float64 value
+# ---
+# int32 CODE_NEGATIVE_VALUE = -110
+# std_msgs/Result result
+# float64 root
+#
+# If a negative value is provided, the server could return the generic CODE_INVALID_ARGUMENT
+# as result.code, or it could provide more context with CODE_NEGATIVE_VALUE.
+#
+# As a general guideline,
+# Positive values are successes
+# Negative values are failures
+#       Values between -1 and -100 are for some generic high-level failures
+#       Values between -101 and -1000 are for client (caller) errors (a la HTTP 4XX errors)
+#       Values between -1001 and -2000 are for server errors (a la HTTP 5XX errors)
 
-uint8 RESULT_FEATURE_UNSUPPORTED   = 0    # Feature is not supported by the simulator, check GetSimulatorFeatures.
-                                          # While feature support can sometimes be deduced from presence of corresponding
-                                          # service / action interface, in other cases it is about supporting certain
-                                          # call parameters, formats and options within interface call.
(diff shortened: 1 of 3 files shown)


mjcarroll:
Added to PMC agenda for May 6th.

std_msgs/msg/Result.msg
```diff
@@ -0,0 +1,18 @@
...
+uint8 RESULT_NOT_FOUND             = 2    # No match for input (such as when entity name does not exist).
+uint8 RESULT_INCORRECT_STATE       = 3    # Simulator is in an incorrect state for this interface call, e.g. a service
+                                          # requires paused state but the simulator is not paused.
+uint8 RESULT_OPERATION_FAILED      = 4    # Request could not be completed successfully even though feature is supported
+                                          # and the input is correct; check error_message for details.
+                                          # Implementation rule: check extended codes for called service
+                                          #  (e.g. SpawnEntity) to provide a return code which is more specific.
+
+uint8 result                              # Result to be checked on return from service interface call
+string error_message                      # Additional error description when useful.
```
fujitatomoya, line 41: result and constants looks simulator specific, i do not think those are generic service response message in system layer. btw, do we not have something like `error_message` description when the service server returns the error? if that is missing, i think that would be useful for system? (user can leave it empty if they do not use it?)

tfoote:
Overall I think that adding this sort of message is a good idea. We'll need to evolve this to be slightly more generic than the current simluation specific descriptions but we've seen similar patterns in other places like nav2 as well


https://github.com/ros-navigation/navigation2/blob/4c8e23cb0ff391dce1f9d2fa00fda2f2cd36f27b/nav2_msgs/srv/LoadMap.srv#L6-L15 

With the goal of not reinventing the wheel. As we look to make this slightly more canonical there's a relatively well established set of error codes used in Abseil and gRPC

https://abseil.io/docs/cpp/guides/status-codes
https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto

As a first suggestion I might suggest that we leverage this existing standard instead of making a new one.

ros-discourse:

https://discourse.ros.org/t/ros-pmc-minutes-for-may-6-2025/43576/1

gavanderhoorn:
Wanted to add a *textual* +1: I've used a similar message/structure in *many* projects, especially when error recovery is (at least partially) delegated to/implemented by machines.

It also aligns with how many RPC systems work, especially in more traditional and established fields.

re: fields: error messages are nice, for humans.

Error codes make things like state machines and behaviour trees -- and the humans implementing/modelling them -- much happier.

(although perhaps with LLMs we could argue error messages could also be used now, but that doesn't mean we can't cater to both)

The current set of constants is of course specific to what was discussed in ros-simulation/simulation_interfaces#1, but that can be changed and should not be a blocker here.

re: constant values: I second @tfoote's suggestion to try and reuse an existing specification -- although perhaps not necessarily limiting candidates to two Google projects.

I would also suggest making `result` a signed int, allowing for negative numbers, and making it 32bit (at least) to allow sufficient space for defining user/OEM-defined status/error messages, in addition to reserved values defined by the message/spec itself.

Could even be nice to introduce categories akin to HTTP response codes (`2xx`, `3xx`, etc). Grouping can make handling errors even more convenient, and standardised groups should allow reuse of code, or at least of experience/best-practices.

std_msgs/msg/Result.msg
```diff
@@ -0,0 +1,18 @@
...
+uint8 RESULT_OK                    = 1
+uint8 RESULT_NOT_FOUND             = 2    # No match for input (such as when entity name does not exist).
+uint8 RESULT_INCORRECT_STATE       = 3    # Simulator is in an incorrect state for this interface call, e.g. a service
+                                          # requires paused state but the simulator is not paused.
+uint8 RESULT_OPERATION_FAILED      = 4    # Request could not be completed successfully even though feature is supported
+                                          # and the input is correct; check error_message for details.
+                                          # Implementation rule: check extended codes for called service
+                                          #  (e.g. SpawnEntity) to provide a return code which is more specific.
+
+uint8 result                              # Result to be checked on return from service interface call
```
gavanderhoorn: See my https://github.com/ros2/common_interfaces/pull/279#issuecomment-2857400082.

I'd suggest:

```suggestion
int32 result                              # Result to be checked on return from service interface call
```

signed to allow negative values to indicate failure in general, and 32 bit to provide sufficient space for reserved ranges, predefined and user-defined values (and it's likely padding / alignment requirements will make this field 4 byte in any case).

mjcarroll:
> Could even be nice to introduce categories akin to HTTP response codes (2xx, 3xx, etc). Grouping can make handling errors even more convenient, and standardised groups should allow reuse of code, or at least of experience/best-practices.

I will note that the `rpc.proto` codes also have equivalent HTTP mappings, which is helpful for translating between ecosystems.

std_msgs/msg/Result.msg
mjcarroll: > user-defined values

I think a header similar to `image_encodings.hpp` in sensor_msgs may be useful here for managing canonical entries versus user-defined.

std_msgs/msg/Result.msg
DLu: > result and constants looks simulator specific

As discussed in the PMC, I copied it over "as-is". It's updated now. 

> btw, do we not have something like `error_message` description when the service server returns the error? i

That existed in ROS 1 but cannot exist in ROS 2 because of the DDS spec (someone else can probably back it up with more precise language)

DLu:
I have taken the suggestions, and rewritten the message with
* less simulator specifics
* `int32` everywhere instead `uint8`
* Multiple "classes" of errors
* Inspiration from existing standards. 

Generally, I was using the [`code.proto`](https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto) definition as a guideline. However, I didn't include every value. Here's what I left off. 
 * `UNAUTHENTICATED` seemed too similar to `PERMISSION_DENIED`
 * `RESOURCE_EXHAUSTED` seemed unclear in this context
 * `ABORTED` seemed to similar to `CANCELLED`
 * `DATA_LOSS` didn't seem to apply here. 



> Could even be nice to introduce categories akin to HTTP response codes (`2xx`, `3xx`, etc).

I now have two separate blocks for Client and Server errors. I would have liked to make it so that there was a particular bit set for each, but in the negative number space, I feel like that takes more mental load than its worth to just do a range check.

zakmat:
Will this promotion be a part of the upcoming Kilted release?

DLu:
> Will this promotion be a part of the upcoming Kilted release?

No, this is being planned for a future release.

PR #281, opened 2025-05-03
WIP: Added BaseTrajectory messages definition in trajectory_msgs for mobile base trajectory following

ahsanyusob:
Hi,

I’m proposing two new messages, `BaseTrajectory` and `BaseTrajectoryPoint`, to the `trajectory_msgs` package. These are designed for planning the trajectory of mobile robot bases, rather than joints.

- **`BaseTrajectory`**: A sequence of trajectory points that the robot base should follow over time.
- **`BaseTrajectoryPoint`**: A single point in the trajectory, specifying the target pose, velocity, and acceleration.

These should be useful for trajectory or path-following controllers in mobile robots. Let me know what you think!

### DIFF
--- a/trajectory_msgs/CMakeLists.txt
+++ b/trajectory_msgs/CMakeLists.txt
@@ -18,6 +18,8 @@ find_package(rosidl_default_generators REQUIRED)
 find_package(std_msgs REQUIRED)
 
 set(msg_files
+  "msg/BaseTrajectory.msg"
+  "msg/BaseTrajectoryPoint.msg"
   "msg/JointTrajectory.msg"
   "msg/JointTrajectoryPoint.msg"
   "msg/MultiDOFJointTrajectory.msg"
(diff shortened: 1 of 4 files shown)


tfoote (changes requested):
Can you explain what you're tying to do differently than the MultiDOFJointTrajectory? This appears to be basically a new version of that except that it can only capture one "joint" at a time.

And you've added names for points along the trajectory. Can you explain why you're adding names to all the points? Trajectories are typically quite dense without names for each point along the trajectory as they're rapidly sampled in time.

ahsanyusob:
Hi, I’m looking for something similar to nav_msgs/Path to define a mobile base trajectory relative to the map, but with stamped desired speed (and acceleration) at each trajectory point. It’s similar to MultiDOFJointTrajectory, except that it’s intended for a single rigid body or mass point moving in local 3D space.

You are right about the name. Initially, I considered linking each trajectory point name to a waypoint/node from the topological map, but as I worked further on it, I realized that this doesn’t make sense, since the trajectory doesn’t necessarily follow the nodes exactly.

PR #287, opened 2025-06-24
add mesh vertex colours and normals

christian-rauch:
## Description

Add vertex colours and normals to the mesh message definition.

Well, the new fields are optional, but the message hash changes, thus rendering previously stored messages (bag files) invalid.

No. I know how to program :-)

### Additional Information

See https://www.open3d.org/docs/latest/python_api/open3d.geometry.TriangleMesh.html for examples of what additional information a mesh may contain and https://github.com/naturerobots/mesh_tools/tree/main/mesh_msgs for prior art.

### DIFF
--- a/shape_msgs/CMakeLists.txt
+++ b/shape_msgs/CMakeLists.txt
@@ -13,6 +13,7 @@ endif()
 
 find_package(ament_cmake REQUIRED)
 find_package(geometry_msgs REQUIRED)
+find_package(std_msgs REQUIRED)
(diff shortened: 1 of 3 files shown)


ahcorde (approved):

ahcorde (changes requested):
we discussed this in the ROS 2 PMC meeting today, and we think that we can add a new message called: `ColoredMesh`, `ColorizedMesh` or `MeshWithColor`. To avoid the issue with bag files

PR #297, opened 2025-09-19
Set of relevant messages for embedded sensors.

bperseghetti:
## Description

Adds a bounded string header useful for static allocation on embedded platforms. IMU, magnetometer and pressure sensors added as they exist in real world. Note an IMU is not an Absolute Orientation Sensor. Introduces a Vector3 version of the Point32.


tfoote (changes requested):
We had a chance to talk about this at the ROS PMC and at a high level we don't think that this is a good approach for us to move the community forward towards supporting embedded platforms. This will start us fragmenting the messages used in the community.

In particular for the new Header we'd rather consider adjusting the header to have a max size overall through a REP process which would enable us to consider picking a maximum size for the messages to allow preallocation.

However this would likely end up with a much bigger maximum size than smaller embedded use cases like it seems that you want to do. 

A longer term approach talked was to extend the ability of the type support generator to enable a compile time flag which can set maximum field lengths that someone compiling for a custom deployment can reduce the maximum size lower and if a message is recieved with the fields too long it will get rejected. This will allow those with constrained systems to tune it for their deployment, but not cause ecosystem fragmentation in the same way as trying to create parallel messages and then tools will need logic to process either or both.

ros-discourse:

https://discourse.openrobotics.org/t/ros-pmc-minutes-for-september-30-2025/50379/1

PR #304, opened 2025-11-07
Add Altitude.msg

### LINKED ISSUE #294: [sensor_msgs] Proposal: Add Altitude.msg (scalar vertical position + variance)
### Description

Add a new scalar message to sensor_msgs named Altitude.msg, containing a timestamped vertical position z in meters (positive up), with an optional uncertainty. This fills the gap between existing messages that either (a) embed altitude together with latitude/longitude (NavSatFix) or (b) represent barometric pressure or generic ranges rather than altitude itself. The proposed definition mirrors existing scalar messages (e.g., Temperature, Illuminance) that carry a variance field.

### Motivation

Many robotics applications need a standardized, timestamped scalar for vertical position (z) that is independent of latitude/longitude. The most common sources are barometric altimeters (converting absolute air pressure to altitude using a reference such as QNH or a locally defined “home” level), radar/laser/ultrasonic altimeters that directly measure height above ground along a downward beam (AGL), and underwater pressure sensors that map hydrostatic pressure to depth; DVLs and single‑beam echosounders also report altitude above the seabed as a single range. 
Today these signals are typically published as FluidPressure, NavSatFix, or Range (or as community‑specific messages), which forces consumers to infer semantics, perform conversions, or carry unrelated fields. A minimal Altitude.msg, Header + altitude [m] (positive up, negative for depth) + variance [m²], would let drivers and estimators exchange vertical‑only observations with unambiguous units, frames, and uncertainty, reducing boilerplate conversions and improving interoperability in state‑estimation pipelines.

Why not Range? Range denotes along-beam distance from an active ranger and has different semantics/fields (min/max, FOV, radiation type). Many altitude sources are not active rangers (e.g., barometers), and even when they are (radar altimeters), the semantic intent is “vertical position”, not raw range.

### Design / Implementation Considerations

Units & Sign (REP-103): meters, +Z up; negative altitude = depth (explicit in docstring). This keeps consistency with ROS axis conventions. 

Uncertainty: follow scalar pattern → variance (float64, m²), where 0.0 = unknown, as in Temperature/FluidPressure/Illuminance.

Frame semantics (REP-105): header.frame_id declares the measurement frame (e.g., base_link, imu_link, map, or earth).

### Additional Information

I’m opening this issue to quickly gauge community feedback on adding a minimal, source‑agnostic vertical position message to sensor_msgs; if there is support, I volunteer to author the PR (message and docs)

ViktorCVS:
## Description

Add a new scalar message to sensor_msgs named Altitude.msg, containing a timestamped vertical position z in meters (positive up), with an optional uncertainty. This fills the gap between existing messages that either (a) embed altitude together with latitude/longitude (NavSatFix) or (b) represent barometric pressure or generic ranges rather than altitude itself. The proposed definition mirrors existing scalar messages (e.g., Temperature, Illuminance) that carry a variance field.

Fixes #294 

No

No

### Additional Information

**Intended publishers (examples)**
- Barometric altimeters → `reference=REFERENCE_MSL` (when the sensor/system provides MSL-referenced altitude), or `reference=REFERENCE_LOCAL_DATUM` (when referenced to a publisher-defined local vertical datum).
- GNSS receivers → `reference=REFERENCE_WGS84` (altitude above the WGS84 ellipsoid; matches NavSatFix semantics), or `reference=REFERENCE_EGM96` (if configured to output altitude relative to the EGM96 geoid model).
- Underwater pressure sensors (depth) → `reference=REFERENCE_SURFACE`; report depth as negative altitude.

Before approving or merging, @tfoote requested additional files. I’ll address this as soon as possible, but I won’t have time over the next few weeks. You can see the request [here](https://github.com/ros2/common_interfaces/issues/294#issuecomment-3379003533)

### DIFF
--- a/sensor_msgs/msg/Altitude.msg
+++ b/sensor_msgs/msg/Altitude.msg
@@ -1,41 +1,34 @@
 # Single vertical position (altitude/depth) measurement.
 #
-# Units & sign (REP-103): meters, +Z up. Negative altitude indicates depth
-# below the chosen reference.
+# Units & sign (REP-103): meters, +Z up. Negative altitude indicates depth.
 #
 # Semantics (REP-105):
-# - 'altitude' is the vertical position along the +Z axis of header.frame_id.
-# - header.frame_id is the frame in which the measurement is expressed
-#   (e.g., base_link, imu_link, map, earth).
+# - 'altitude' is the height (or depth if negative) of the origin of header.frame_id,
+#   measured vertically in the selected 'reference'.
+# - header.frame_id identifies the point this measurement applies to.
 #
 # Reference:
-# The 'reference' field indicates what this measurement is relative to:
-#   REFERENCE_UNKNOWN (0): unspecified or mixed semantics.
-#   REFERENCE_DATUM   (1): relative to a vertical datum / ellipsoid / geoid.
-#                         When used, 'datum' should name the model, e.g.
-#                         "WGS84", "EGM96", "MSL", or a local datum identifier.
-#   REFERENCE_SURFACE (2): relative to a local physical surface (ground, seabed,
-#                         water surface). Typical for AGL/height-above-surface sensors.
+# The 'reference' field indicates what this measurement is relative to.
 #
 # Uncertainty:
 # - 'variance' is the measurement variance in m^2; 0.0 indicates unknown.
 
 # Reference enumeration
 uint8 REFERENCE_UNKNOWN=0
-uint8 REFERENCE_DATUM=1
-uint8 REFERENCE_SURFACE=2
+uint8 REFERENCE_MSL=1            # altitude above mean sea level (as provided by the sensor/system)
+uint8 REFERENCE_WGS84=2          # altitude above WGS84 ellipsoid (matches NavSatFix semantics)
+uint8 REFERENCE_EGM96=3          # altitude relative to EGM96 geoid model
+uint8 REFERENCE_LOCAL_DATUM=4    # altitude relative to a publisher-defined local vertical datum
+uint8 REFERENCE_SURFACE=5        # altitude relative to a local physical surface (ground, seabed, water surface, etc.)
 
-# Timestamp and frame of the measurement
-std_msgs/Header header        # timestamp is when the altitude was measured;
-                              # frame_id is the frame whose +Z defines "up"
+# Timestamp and point of the measurement
+std_msgs/Header header  # stamp: measurement time; frame_id: measurement point
 
 # Scalar vertical position
-float64 altitude              # Vertical position in meters (+Z up); negative for depth
+float64 altitude        # Vertical position in meters (+Z up); negative for depth
 
 # Measurement uncertainty
-float64 variance              # 0 is interpreted as variance unknown
+float64 variance        # 0 is interpreted as variance unknown
 
-# Optional name of the datum/surface
-string datum                  # Name of vertical datum or surface if applicable
-                              # (e.g., "WGS84", "EGM96", "MSL", "GROUND", "SEA_SURFACE");
-                              # empty if unknown or not applicable
+# Which reference 'altitude' is relative to (see REFERENCE_* constants above)
+uint8 reference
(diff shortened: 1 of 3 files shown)


sensor_msgs/msg/Altitude.msg
```diff
@@ -0,0 +1,41 @@
...
+                              # frame_id is the frame whose +Z defines "up"
+
+# Scalar vertical position
+float64 altitude              # Vertical position in meters (+Z up); negative for depth
+
+# Measurement uncertainty
+float64 variance              # 0 is interpreted as variance unknown
+
+# Optional name of the datum/surface
+string datum                  # Name of vertical datum or surface if applicable
```
tfoote: Each datum supported should be in the enum above, not a string here.

sensor_msgs/msg/Altitude.msg
```diff
@@ -0,0 +1,41 @@
...
+# Units & sign (REP-103): meters, +Z up. Negative altitude indicates depth
+# below the chosen reference.
+#
+# Semantics (REP-105):
+# - 'altitude' is the vertical position along the +Z axis of header.frame_id.
+# - header.frame_id is the frame in which the measurement is expressed
+#   (e.g., base_link, imu_link, map, earth).
+#
+# Reference:
+# The 'reference' field indicates what this measurement is relative to:
```
tfoote: This field doesn't appear to exist.

sensor_msgs/msg/Altitude.msg
```diff
@@ -0,0 +1,41 @@
+# Single vertical position (altitude/depth) measurement.
+#
+# Units & sign (REP-103): meters, +Z up. Negative altitude indicates depth
+# below the chosen reference.
+#
+# Semantics (REP-105):
+# - 'altitude' is the vertical position along the +Z axis of header.frame_id.
+# - header.frame_id is the frame in which the measurement is expressed
+#   (e.g., base_link, imu_link, map, earth).
```
tfoote: We need to be careful on this. base_link and imu_link are rarely actually vertical in the world. Isn't this about being vertical in the datum?

tfoote (changes requested):
A few inline comments. 

And from my previous issue comments 

> If you can prototype of the converter which can convert all the different datatypes into the common one. And provide a demonstration of one of the state estimators or other potential downstream user that can demonstrate that the proposed message covers the needs on both sides.

Before we land something in common_interfaces we want to see implementations and real use cases that are working to know that we have coverage of the needs. When a proposal hits realitiy there's often an iteration on the design required. And once it's here changes are very expensive to make due to our stability goals.

sensor_msgs/msg/Altitude.msg
ViktorCVS: Thanks! It was addressed in commit e9adbda4.

sensor_msgs/msg/Altitude.msg
ViktorCVS: Sorry, I mistakenly cut off the end of the file. It was addressed in commit https://github.com/ros2/common_interfaces/commit/e9adbda441a08778e79dfdeeac0d8a9a05cd60c9.

sensor_msgs/msg/Altitude.msg
ViktorCVS: Thanks! It was addressed in commit https://github.com/ros2/common_interfaces/commit/e9adbda441a08778e79dfdeeac0d8a9a05cd60c9.

ViktorCVS:
> A few inline comments.
> 
> And from my previous issue comments
> 
> > If you can prototype of the converter which can convert all the different datatypes into the common one. And provide a demonstration of one of the state estimators or other potential downstream user that can demonstrate that the proposed message covers the needs on both sides.
> 
> Before we land something in common_interfaces we want to see implementations and real use cases that are working to know that we have coverage of the needs. When a proposal hits realitiy there's often an iteration on the design required. And once it's here changes are very expensive to make due to our stability goals.

Thanks again for your review. I’ll address the request from the issue within a week.

PR #329, opened 2026-04-16
add missing msgs for PointWithCovariance

### LINKED ISSUE #328: Proposal: Add msgs to support points with covariances
### Description

Currently, the geometry_msgs does only feature the Point and Point32 message to represent a point's position in 3D space. The suggestion would be to add additional messages to include the uncertainty of this position given by the 3D covariance. 

### Motivation

Many applications require the covariance/uncertainty of geometric informations for further processing. E.g. if a detector extracts features/points of some sensor data, it may also give an uncertainty estimation on how good the detection is. This may be used in further processing.

### Design / Implementation Considerations

Add missing msgs:

- PointWithCovariance.msg
- Point32WithCovariance.msg
- PointWithCovarianceStamped.msg
- Point32WithCovarianceStamped.msg
- (Point32Stamped.msg - missing in existing representations, adding for completeness)

### Additional Information

_No response_

mleitenstern:
## Description

This PR adds additional messages to the package to represent a point's position in 3D-space with a 3D covariance.
The added msgs support 32-bit and 64-bit precision points and are available with or without a header. A more detailed explanation on the necessity and exemplary usage of these msg-types can be found in #328.
In summary, the following msgs are added:

- Point32WithCovariance.msg
- Point32WithCovarianceStamped.msg
- PointWithCovariance.msg
- PointWithCovarianceStamped.msg

Additionally, the missing msg Point32Stamped was added to ensure completeness. 
Fixes #328 

Yes, users will have access to additional msg-types to represent points with additional covariances. However, all existing msgs are kept unmodified.

Yes, Claude Code using Opus 4.6


geometry_msgs/msg/Point32Stamped.msg
```diff
@@ -0,0 +1,9 @@
+# This represents a Point with reference coordinate frame and timestamp
+# (with 32 bits of precision).
+# It is recommended to use PointStamped wherever possible instead of
+# Point32Stamped.
+#
+# This recommendation is to promote interoperability.
+
+std_msgs/Header header
+Point32 point
```
tfoote: This doesn't have semantic meaning on it's own so shouldn't be defined here.

geometry_msgs/msg/Point32WithCovariance.msg
```diff
@@ -0,0 +1,15 @@
+# This represents a point in free space with uncertainty (with 32 bits of precision).
+# It is recommended to use PointWithCovariance wherever possible instead of
+# Point32WithCovariance.
+#
+# This recommendation is to promote interoperability.
+#
+# This message is designed to take up less space when sending
+# lots of points at once, as in the case of a PointCloud.
+
+Point32 point
```
tfoote: Point32 is focused on being used in areas where Point (64bit) is too much data. Optimizing 32bits here doesn't make sense when you're attaching 288 below.

geometry_msgs/msg/Point32WithCovariance.msg
```diff
@@ -0,0 +1,15 @@
...
+
+# Row-major representation of the 3x3 covariance matrix
+# In order, the parameters are:
+# (x, y, z)
+float32[9] covariance
```
tfoote: Why is this float32? Every other instance of covariance is float64 which is our standard representation as per REP 103: https://reps.openrobotics.org/rep-0103/

tfoote (changes requested):
As per your self filed ticket these all seem to be "adding for completeness" which is not how or why we should be adding messages. They dilute the standard messages which should have concrete use cases and strong semantic meaning. 

Before we standardize messages we want to see compelling use cases which demonstrate the value of the messages.

mleitenstern:
@tfoote i included use cases of the proposed messages in the corresponding issue #328. 
Regarding your comments regarding the proposed changes: 
I agree that the standard representation for covariances is float64. Would your suggestion be to only integrate PointWithCovariance and PointWithCovarianceStamped and exclude the corresponding 32-bit representations?

geometry_msgs/msg/Point32Stamped.msg
mleitenstern: what do you mean by "semantic meaning"?

tfoote: What does this represent?  A stamped datatype should be able to be interpreted without extra information as it's designed to be sent on it's own. 

Is it an obstacle, or a goal? You need to treat those very differently and don't want to accidentally cross those wires when navigating.

tfoote (commented):
We use 64bit datatypes by default to avoid fragmentation of messages. To use the 32bit representation please justify why we should create a duplicate version of the message which will cause fragmentation of all downstream users who would need to potentially subscribe and deduplicate parallel streams representing the same data.

geometry_msgs/msg/Point32Stamped.msg
mleitenstern: i understand, but shouldn't this package just contain geometric primitives?
PointStamped does also not have any semantic meaning on its own, but this comes from composing it within other messages or topic naming.
Please correct me if i'm wrong.

mleitenstern:
> We use 64bit datatypes by default to avoid fragmentation of messages. To use the 32bit representation please justify why we should create a duplicate version of the message which will cause fragmentation of all downstream users who would need to potentially subscribe and deduplicate parallel streams representing the same data.

That's a fair point, and fragmentation is a stronger argument than the memory savings I had in mind. My thought was that downstream messages composing arrays of PointWithCovariance (e.g. polylines or large landmark sets) could become sizeable, and Point32 already exists for that reason in the point-cloud context — so I included the 32-bit variants for parity. But I don't have measurements showing the 64-bit version is actually a problem in practice, and the cost of two parallel streams across the system is definitely outweighing it).

Would you agree with dropping the Point32WithCovariance and Point32WithCovarianceStamped variants from the proposal?
If a concrete bandwidth or memory issue arises later in a real deployment, the 32-bit version can be revisited as a separate proposal with actual numbers to back it up.
That also simplifies the question of Point32Stamped — I'd included it for completeness, but if we're not adding 32-bit covariance variants, it's probably better left out too and addressed separately if there's demand.

tfoote (approved):
Thanks for iterating. I know this is less scope now but I think the more measured approach will be better overall. 

I expect this to be a useful submessage for people to be able to build semantically meaningful datatypes. 

For the example use case see: https://github.com/ros2/common_interfaces/issues/328#issuecomment-4505680077

mleitenstern:
@tfoote am i expected/allowed to tick this box or are you going to take care of it ? :)

PR #332, opened 2026-05-26
fix whitespace in CameraInfo

christian-rauch:
## Description

This fixes a few whitespace issues in the ´CameraInfo.msg` to align comments. See the ROS 1 definition at https://github.com/ros/common_msgs/blob/noetic-devel/sensor_msgs/msg/CameraInfo.msg for comparison.


mjcarroll:
Since this is purely a whitespace/comment change it will have no impact on the rosidl pipeline.

mjcarroll (approved):

PR #338, opened 2026-06-19
std_srvs: add SetFloat32, SetInt32, SetString, SetUInt8, SetUInt8Array services

xbroquere:
## Summary

Add SetFloat32, SetInt32, SetString, SetUInt8, and SetUInt8Array service files.

## Description

This PR extends std_srvs with five additional setter-style services and wires them into interface generation.

Changes:
- Add new service definitions:
  - std_srvs/srv/SetFloat32.srv
  - std_srvs/srv/SetInt32.srv
  - std_srvs/srv/SetString.srv
  - std_srvs/srv/SetUInt8.srv
  - std_srvs/srv/SetUInt8Array.srv

- Update std_srvs/CMakeLists.txt to generate interfaces for all new services.
- Update std_srvs/README.md to document the new services with usage-oriented descriptions.
- Update std_srvs/CHANGELOG.rst under Forthcoming.

Yes.
New service interface types are now available in std_srvs for package users. Existing services are unchanged.

Yes.
Tool: GitHub Copilot
Model: GPT-5.3-Codex
Usage: Assisted with writing the PR description

### Additional Information

- Change is additive and non-breaking for existing std_srvs users.
- Verified in local workspace with colcon build for std_srvs and a dedicated test package exercising all new services.

### DIFF
--- a/std_srvs/CHANGELOG.rst
+++ b/std_srvs/CHANGELOG.rst
@@ -2,6 +2,12 @@
 Changelog for package std_srvs
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+Forthcoming
+-------------------
+* Add additional std_srvs service definitions:
+  SetFloat32, SetInt32, SetString, SetUInt8, and SetUInt8Array.
+* Contributors: Xavier BROQUERE
+
 5.10.0 (2026-05-06)
 -------------------
 
--- a/std_srvs/CMakeLists.txt
+++ b/std_srvs/CMakeLists.txt
@@ -17,6 +17,11 @@ find_package(rosidl_default_generators REQUIRED)
 set(srv_files
   "srv/Empty.srv"
   "srv/SetBool.srv"
+  "srv/SetFloat32.srv"
+  "srv/SetInt32.srv"
+  "srv/SetString.srv"
+  "srv/SetUInt8.srv"
+  "srv/SetUInt8Array.srv"
   "srv/Trigger.srv"
 )
 rosidl_generate_interfaces(${PROJECT_NAME}
--- a/std_srvs/README.md
+++ b/std_srvs/README.md
@@ -7,6 +7,11 @@ For more information about ROS 2 interfaces, see [docs.ros.org](https://docs.ros
 ## Services (.srv)
 * [Empty.srv](srv/Empty.srv): A service containing an empty request and response.
 * [SetBool.srv](srv/SetBool.srv): Service to set a boolean state to true or false, for enabling or disabling hardware for example.
+* [SetFloat32.srv](srv/SetFloat32.srv): Service to set a float32 value, for setting a numeric threshold or gain for example.
+* [SetInt32.srv](srv/SetInt32.srv): Service to set an int32 value, for setting an integer mode or counter for example.
+* [SetString.srv](srv/SetString.srv): Service to set a string value, for selecting a profile or a configuration name for example.
+* [SetUInt8.srv](srv/SetUInt8.srv): Service to set an unsigned 8-bit integer value, for setting a small command code for example.
+* [SetUInt8Array.srv](srv/SetUInt8Array.srv): Service to set an array of unsigned 8-bit integer values, for sending a compact payload for example.
 * [Trigger.srv](srv/Trigger.srv): Service with an empty request header used for triggering the activation or start of a service.
 
 
--- a/std_srvs/srv/SetFloat32.srv
+++ b/std_srvs/srv/SetFloat32.srv
@@ -1,0 +1,4 @@
+float32 data # e.g. for setting a numeric threshold
+---
+bool success   # indicate successful run of triggered service
+string message # informational, e.g. for error messages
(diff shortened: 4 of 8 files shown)


asymingt:
Hi @xbroquer. First off, thank you for the contribution. We have to be very careful with changes like these because they set a precedent about what should be accepted into the core message libraries. To be a good fit here, we have to make sure that the usefulness brought to the wider community outweighs the additional compilation time.

For this reason I need to know a few things about why you added these messages.

1. All ROS nodes have a parameter mechanism, which allows other entities to get and set values. I have a hunch that this mechanism might be what you are after. Here's an introduction to them: https://docs.ros.org/en/lyrical/Tutorials/Beginner-CLI-Tools/Understanding-ROS2-Parameters/Understanding-ROS2-Parameters.html.
2. Assuming that the parameter mechanism doesn't work for you, can you justify why you aren't able to create and use your own custom messages in you own custom ROS package to achieve what you need.  See: https://docs.ros.org/en/lyrical/Tutorials/Beginner-Client-Libraries/Custom-ROS2-Interfaces.html

If you still think that these messages are useful in some way that I haven't covered in the two points above, can you prepare a justification and I will bring this PR to the ROS PMC for discussion.

xbroquere:
Hi @asymingt  ,

Thank you for the review. I agree this change requires strong justification as it impacts core interfaces.

Context: I am a software engineer at Renault Group, leading a ROS-based platform for ADAS prototyping. We started on ROS 1 Kinetic, then migrated through each ROS 2 LTS generation. We are currently on Jazzy and have already begun supporting Lyrical. We use ROS 2 parameters extensively for configuration and runtime algorithm tuning.

This proposal is not intended to replace parameters, but to address a different interface contract.

Rationale:
Parameters are suited for node configuration and state, while Set* services are for immediate command calls with typed input and direct feedback (success/message). Parameters remain the default for configuration and tuning, and Set* services are only used for command-style RPC interactions.

Why parameters are not sufficient for this subset of interactions:
1. Some interactions are commands, not configuration updates.
2. We need immediate per-call feedback (success/failure + message), not just changed state visibility.

Concrete examples:
1. SetUInt8Array: gateway unlock/control where the client sends a byte payload and gets immediate success/message feedback.
2. SetString: command inputs such as profile or mode selection, or file-load requests, where the result must be returned synchronously.
3. SetUInt8: compact command APIs where the value represents an enum-like command (e.g., start/stop/mode/action selection) and the caller needs immediate feedback that the command was accepted/executed (bool + message).
4. SetInt32: diagnostic command APIs where the int32 request carries a numeric identifier (e.g., a DID in automotive DoIP diagnostics) and the service returns immediate success/message feedback containing the result or diagnostic error.

Why upstream instead of keeping private duplicates:
1. We already maintain an internal xxx_std_srvs package extending std_srvs with generic Set* primitives.
2. These interfaces are not domain-specific and represent recurring patterns duplicated across packages.
3. This proposal extends existing primitives (e.g., Empty, SetBool, and Trigger).
4. Upstreaming avoids repeated redefinition of identical simple RPC interfaces (primitive payload + bool/message response).
5. As standardized interfaces, these services can improve interoperability and benefit the wider ROS community.

Optional scope note: SetFloat32 is included for completeness of scalar primitives, but can be omitted if needed to keep the scope minimal.

asymingt:
Thank you for taking time to respond with a good amount of detail, and I understand why parameters aren't a good fit for your problem. I'll raise your PR at the ROS PMC meeting next Tuesday, and I'll get back to you with next steps shortly therefter. 

In the interim, is there any open source project that you can point to that shows a good use case for these additional service types. We often get asked by PMC members to see these when discussing additions to the core interfaces.

xbroquere:
Thank you for the follow-up.
I don't have a public open-source example today, as the current use cases come from internal projects.
However, I plan to open-source some generic tooling soon, including a fork of record_ros and an rqt_reconfigure-like application. Both would be good candidates to demonstrate practical use cases for these additional service types.

gavanderhoorn:
As a bit of a drive-by comment almost: previous proposals for adding the kind of service types contributed here were not accepted on the grounds of the message types being too generic.

To take the `SetInt32` example: a more appropriate way to model this interaction could be to design a more semantically meaningful service message and create servers and clients using that message. Perhaps something like `SetDID` as a type (I'm not an expert here, so just to give it a name), or perhaps even more specific (`Trigger<Something>` or `Read<SomeSensorData>`).

This to not have to rely on topic/service/action name to convey the semantics of the interaction (ie: "*diagnostic command APIs where the int32 request carries a numeric identifier (e.g., a DID in automotive DoIP diagnostics)*") -- essentially avoiding the situation where *any* client that knows how to create a `SetInt32` request could interact with the server, even if the request makes no sense.

The downside is of course less reusability of clients and servers (among other things), but IIRC, that was accepted in the past as the benefits of clearer semantics were expected to outweigh (some of) those downsides.

asymingt:
Hi @xbroquer - I raised the issue in yesterday's (2026-07-07) ROS PMC meeting. The sentiment of the group was largely aligned with the comments from @gavanderhoorn. The overarching feedback was that the Set<X> semantics are too generic, and that merging PRs like this sets a precedent that's hard to maintain.

You current approach of providing `xxx_std_srvs` packages shared across your domains (through an underlay or shared package) is indeed the right way to achieve what you need.

Thank you again for putting up a PR with the goal of improving ROS. I realize that you might find it disappointing that we aren't able to accept your contribution, and that we need to close this PR. Decisions like these are hard because they involve real trade-offs in the community.

